PR #55, opened 2026-04-06
ci: add clang-format check on changed files (S4-P6)

AlvinYangZF:
## Summary
- `.clang-format`: LLVM-based config (4-space indent, 120 col, Linux braces)
- `.github/workflows/format-check.yml`: per-PR check on **changed files only** — no full-codebase reformat required
- Skips gracefully when no C/H files are modified

## Test plan
- [x] YAML validates
- [ ] GitHub Actions green

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,9 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+TabWidth: 4
+UseTab: Never
+ColumnLimit: 120
+BreakBeforeBraces: Linux
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: false
+SortIncludes: false
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -7,7 +7,6 @@ name: Format Check
 on:
   pull_request:
     branches: [master, main]
-  workflow_dispatch:
 
 jobs:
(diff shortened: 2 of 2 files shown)


AlvinYangZF:
I re-checked the latest revision.

The issue I called out earlier is resolved from my side: the workflow no longer advertises a `workflow_dispatch` path that cannot compute a valid diff base. Keeping this as a PR-only changed-files check is the cleaner shape for the current implementation.

At this point I do not see a remaining blocker in this PR.

yangzf_codex

AlvinYangZF:
One issue I found in the current `format-check.yml` shape: it can false-fail on deleted C/H files.

The workflow builds `CHANGED` with:

```bash
git diff --name-only "$BASE" "$HEAD" -- '*.c' '*.h'
```

and then passes that list straight into `clang-format --dry-run -Werror`.

That means a PR that deletes `foo.h` will still include `foo.h` in `CHANGED`, but the file no longer exists in the checkout, so `clang-format` fails with `No such file or directory`.

I reproduced that locally with a tiny temp repo: delete a header, run the same `git diff --name-only ... | xargs clang-format --dry-run -Werror`, and the check fails even though there is nothing left to format.

I think this should filter to existing files only (for example via `--diff-filter=d` on the git side, or by checking `[ -f ]` before invoking `clang-format`).

PR #56, opened 2026-04-06
ci: add fio JSON performance metrics parser (S2a-P1)

AlvinYangZF:
## Summary
- `scripts/perf/parse_fio_json.sh`: Extracts IOPS, mean lat, p99 lat from fio JSON output
- `scripts/perf/testdata/sample_fio.json`: Realistic test fixture
- Foundation for Sprint 2 perf regression detection (nightly trend → baseline → ratchet)

## Test plan
- [x] Parser produces valid JSON with correct metrics
- [x] Tested against sample fixture: read 642.5 IOPS / 245.3us mean / 890.1us p99

## Output example
```json
{
  "source": "sample_fio.json",
  "jobs": [{
    "jobname": "randrw_4k",
    "read_iops": 642.5, "write_iops": 275.3,
    "read_lat_mean_us": 245.3, "write_lat_mean_us": 1315.7,
    "read_lat_p99_us": 890.1, "write_lat_p99_us": 3456.2
  }]
}
```


AlvinYangZF:
One issue I found in the current parser behavior: a missing fio `timestamp` is silently turned into a fake historical timestamp instead of being rejected.

Right now the script emits:

```jq
fio_timestamp: (.timestamp // 0 | todate)
```

So if the input JSON has no `timestamp`, the summary still exits successfully and reports:

```json
"fio_timestamp": "1970-01-01T00:00:00Z"
```

For the nightly-trend / baseline use case, that is a dangerous default because it creates a valid-looking but fabricated sample time rather than surfacing bad input.

I reproduced this locally with a minimal fio-like JSON that omitted `timestamp`; the script exited 0 and emitted the Unix epoch.

I think this should either require `.timestamp` to be present and fail otherwise, or emit `null` / a separate validation error instead of coercing to epoch zero.

PR #64, opened 2026-04-09
test: add nightly blackbox cases and frontend coverage report

AlvinYangZF:
## Summary
- **Task 2.2**: `007_nvme_format_nvm.sh` — nightly destructive admin test (write → format → verify erasure)
- **Task 2.3**: `016_fio_nightly_queue_pressure.sh` — nightly queue-pressure fio (4k randrw, iodepth=32, 60s, verify)
- **Task 3.1**: `scripts/coverage/frontend_report.sh` + `make coverage-frontend` — per-module coverage for 7 front-end modules including vhost

Nightly cases are auto-discovered by `blackbox-full` but excluded from PR smoke.

## Test plan
- [x] `make test` passes locally
- [x] Blackbox case scripts are executable and follow existing patterns
- [x] `make help | grep frontend` shows new target
- [ ] CI green on PR

### DIFF
--- a/.coverage-frontend-baseline.json
+++ b/.coverage-frontend-baseline.json
@@ -1,0 +1,12 @@
+{
+  "created_at": "2026-04-08T00:00:00Z",
+  "modules": {
+    "src/vhost/hfsss_nbd_server.c": { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/vhost/nbd_async.c":        { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/pcie/nvme.c":              { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/pcie/nvme_uspace.c":       { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/pcie/queue.c":             { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/controller/shmem_if.c":    { "lines": 0.0, "functions": 0.0, "branches": 0.0 },
+    "src/common/msgqueue.c":        { "lines": 0.0, "functions": 0.0, "branches": 0.0 }
+  }
+}
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,7 @@ COVERAGE_BINS = $(COVERAGE_UT_BINS) $(COVERAGE_E2E_BINS)
 
 # Targets
 .PHONY: all clean directories test systest stress-long help \
-	coverage-build coverage-clean coverage-ut coverage-e2e coverage-merge coverage coverage-selftest \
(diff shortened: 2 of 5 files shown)


AlvinYangZF:
One concern on the new nightly fio case: `016_fio_nightly_queue_pressure.sh` currently combines `--verify=crc32c` with `--norandommap`, `--ioengine=libaio`, `--iodepth=32`, and `--numjobs=2`. That combination weakens the integrity signal rather than strengthening it.

Two reasons:
1. Our own repo already documents this constraint in `docs/PRE_CHECKIN_STANDARD.md`: `numjobs=2` with verify can trigger fio's "multiple writers may overwrite blocks that belong to other jobs" warning and break verify semantics.
2. fio's documentation also warns against using `norandommap` together with verification on async engines at `iodepth > 1`, because it can cause false verification behavior.

On top of that, `hfsss_case_run_fio_json()` only checks `job.error` in the JSON output, so parser/verify warnings of this class do not automatically fail the case.

So as written, this case can report success without providing a reliable end-to-end integrity check. I think this should either switch back to a verify-safe shape (`numjobs=1`, and likely no `norandommap`), or explicitly drop the integrity claim and treat it as a pure queue-pressure stress case.

AlvinYangZF:
One bug in `scripts/coverage/frontend_report.sh`: the script intends to warn and exit 0 when no front-end module coverage is present, but that path is currently unreachable.

Right now it does `lcov --extract "$INFO" ... --output-file "$FRONTEND_INFO"` first, and only afterwards checks `[ ! -s "$FRONTEND_INFO" ]`. If the input tracefile exists and is valid, but contains no matching front-end modules, `lcov --extract` itself exits non-zero with `(empty) no valid records found in tracefile`, so the script aborts before reaching the warning/fallback message.

I reproduced this locally with a minimal valid tracefile that only contained `/tmp/other.c`: `bash scripts/coverage/frontend_report.sh /tmp/pr64-unrelated.info` exited 1 from `lcov` instead of printing the intended "Try using e2e.raw.info instead" guidance.

That means `make coverage-frontend` currently turns "no front-end data yet" into a hard failure. I think this needs `--ignore-errors empty` (or equivalent handling around the extract step) so the script can reach its intended warning path.

AlvinYangZF:
Addressed both comments in d99b9e3:

**1. fio queue-pressure verify safety**

Switched to `numjobs=1` and removed `norandommap`. This avoids the multi-writer verify overlap issue documented in `PRE_CHECKIN_STANDARD.md` and fio's own async-verify caveat. The case now provides a reliable integrity signal at iodepth=32.

**2. frontend_report.sh empty-extract abort**

Added `--ignore-errors empty,empty` and `|| true` to the main `lcov --extract` call. When the input tracefile has no matching front-end modules, the script now reaches the intended warning/exit-0 path instead of aborting.

PR #67, opened 2026-04-10
test: expand system test coverage (DI/NC/PS/WL/GC/PR)

AlvinYangZF:
## Summary
Implements the System Test Plan Phase A — whitebox system tests covering the
NVMe → FTL → NAND path via the \`nvme_uspace_*\` API directly (no QEMU guest needed).

### New test cases

| Category | Cases | File | Status |
|----------|-------|------|--------|
| **Data Integrity** | DI-003, DI-005, DI-008 | systest_data_integrity.c | Added |
| **NVMe Compliance** | NC-011 | systest_nvme_compliance.c | Added |
| **Performance** | PS-001..008 | systest_performance.c (new) | All new |
| **Wear Leveling** | WL-001..005 | systest_wear_gc.c (new) | All new |
| **GC Policy** | GC-001..004 | systest_wear_gc.c (new) | All new |
| **Persistence** | PR-001..007 | systest_persistence.c (new) | All new |

### Supporting changes
- \`src/ftl/wear_level.c\`: implement \`wear_level_update_stats()\` and \`wear_level_check_wear()\` required by WL tests
- \`Makefile\`: add build and run targets for \`systest_performance\`, \`systest_wear_gc\`, \`systest_persistence\`
- Reduced device geometry in new/updated tests to 1ch × 1chip × 32 blocks × 64 pages for fast runs

### Results
- **systest_data_integrity**: 50/50 PASS (6.7 min)
- **systest_nvme_compliance**: 70/70 PASS
- **systest_error_boundary**: 536/536 PASS (untouched)
- **systest_performance**: 23/23 PASS
- **systest_wear_gc**: 44/44 PASS (24 min)
- **systest_persistence**: 39/39 PASS (1 min)
- **Total: 762/762 assertions**, 0 failures

## Test plan
- [x] All 6 systest binaries build clean
- [x] All 762 assertions pass locally
- [x] Fault injection tests (EH-001..008) deferred — require wiring \`fault_check()\` into the HAL data path
- [ ] CI validation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,8 @@ COVERAGE_UT_BINS = $(COVERAGE_BIN_DIR)/test_common $(COVERAGE_BIN_DIR)/test_medi
 	$(COVERAGE_BIN_DIR)/test_msgqueue \
 	$(COVERAGE_BIN_DIR)/test_nvme_admin_cmds \
 	$(COVERAGE_BIN_DIR)/test_nvme_io_cmds \
-	$(COVERAGE_BIN_DIR)/systest_performance
+	$(COVERAGE_BIN_DIR)/systest_performance \
+	$(COVERAGE_BIN_DIR)/systest_persistence
 COVERAGE_E2E_BINS = $(COVERAGE_BIN_DIR)/hfsss-nbd-server
 COVERAGE_BINS = $(COVERAGE_UT_BINS) $(COVERAGE_E2E_BINS)
 
@@ -539,7 +540,7 @@ stress-long: all
 
 # System-level tests (Tier 1)
 .PHONY: systest
-systest: directories $(SYSTEST_DI) $(SYSTEST_NC) $(SYSTEST_EB) $(SYSTEST_PS) $(SYSTEST_WG)
+systest: directories $(SYSTEST_DI) $(SYSTEST_NC) $(SYSTEST_EB) $(SYSTEST_PS) $(SYSTEST_WG) $(SYSTEST_PR)
 	@echo "========================================"
 	@echo "Running system-level tests..."
 	@echo "========================================"
@@ -552,6 +553,8 @@ systest: directories $(SYSTEST_DI) $(SYSTEST_NC) $(SYSTEST_EB) $(SYSTEST_PS) $(S
 	@$(SYSTEST_PS)
 	@echo ""
 	@$(SYSTEST_WG)
+	@echo ""
+	@$(SYSTEST_PR)
 	@echo ""
 	@echo "========================================"
 	@echo "All system-level tests complete!"
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,9 @@ TEST_T10PI = $(BIN_DIR)/test_t10_pi
 SYSTEST_DI = $(BIN_DIR)/systest_data_integrity
 SYSTEST_NC = $(BIN_DIR)/systest_nvme_compliance
 SYSTEST_EB = $(BIN_DIR)/systest_error_boundary
+SYSTEST_PS = $(BIN_DIR)/systest_performance
+SYSTEST_WG = $(BIN_DIR)/systest_wear_gc
+SYSTEST_PR = $(BIN_DIR)/systest_persistence
 TEST_UPLP = $(BIN_DIR)/test_uplp
 TEST_QOS = $(BIN_DIR)/test_qos
 TEST_SECURITY = $(BIN_DIR)/test_security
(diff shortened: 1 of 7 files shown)


AlvinYangZF:
I found a behavioral bug in the new wear-monitoring path: `wear_level_check_wear()` can report a critical alert immediately on a freshly initialized context.

`wear_level_init()` currently zeroes the whole struct and only restores `enabled`, `static_enabled`, `static_threshold`, and `last_static_check_ts`. It does **not** initialize the wear-monitoring fields (`wear_monitoring_enabled`, `wear_alert_threshold`, `wear_critical_threshold`) to the defaults declared in `wear_level.h`.

But `wear_level_check_wear()` then computes:

```c
u32 pct = (ctx->max_erase_count * 100) / max_pe_cycles;
if (pct >= ctx->wear_critical_threshold) ...
if (pct >= ctx->wear_alert_threshold) ...
```

with no guard on `wear_monitoring_enabled`.

So the default thresholds stay at `0`, and the very first call satisfies `pct >= 0`, which turns a fresh context into `WEAR_ALERT_CRITICAL`.

I reproduced this locally with a tiny standalone repro against this PR's `wear_level.c`: after `wear_level_init()`, calling `wear_level_check_wear(..., 1000)` printed `monitor=0 warn=0 crit=0 ... alert=2 criticals=1`.

I think this needs one of:
- initialize the monitoring fields in `wear_level_init()` to the defaults from the header, and/or
- make `wear_level_check_wear()` return `WEAR_ALERT_NONE` when monitoring is disabled.

AlvinYangZF:
The new persistence test target is compiled, but it is not wired into the standard execution paths.

`SYSTEST_PR` and the build rule for `systest_persistence` are present, but:
- `make systest` only runs `DI/NC/EB/PS/WG`
- `coverage-ut` also omits `systest_persistence`

I verified this with `make -n systest` and `make -n coverage-ut` on the PR worktree: neither one emitted any `systest_persistence` command.

So right now the PR adds persistence tests that will build under `all`, but they do not run in the standard systest bundle and do not contribute to the UT coverage flow either. That makes the advertised Phase A coverage expansion incomplete from a CI / regression-protection standpoint.

AlvinYangZF:
A few of the new WL/GC cases are currently informational only, so they do not actually pin down the behaviors the PR says it is covering.

Examples:
- `WL-001`: if PE spread exceeds the target, the test still does `TEST_ASSERT(true)` and passes.
- `WL-002`: if the workload fails to create enough skew to exceed the threshold, the test also `TEST_ASSERT(true)` and passes.
- `GC-003`: the code explicitly says `No strict latency assertion -- observational only` and ends with `TEST_ASSERT(true)`.

That means these cases can never fail on the main condition they are supposed to validate; they only print measurements.

I do not think every performance / WL test needs a razor-tight bound, but as written these particular cases are closer to diagnostics than regression tests. If they are intended to count toward the new system-test coverage claims, I think they need at least one real pass/fail predicate tied to the stated goal of each case.

AlvinYangZF:
Addressed all three review items in 431203d:

**1. `wear_level_init()` bug**
Initialized `wear_monitoring_enabled`, `wear_alert_threshold`, and `wear_critical_threshold` to the header defaults (`WEAR_MONITORING_ENABLED`, `WEAR_ALERT_THRESHOLD=90`, `WEAR_CRITICAL_THRESHOLD=95`). Also added a short-circuit in `wear_level_check_wear()` to return `WEAR_ALERT_NONE` when monitoring is disabled.

**2. `systest_persistence` not wired into execution paths**
Added `$(SYSTEST_PR)` to the `systest` target prereqs and invocation, and added `build-cov/bin/systest_persistence` to `COVERAGE_UT_BINS`. Verified with:
- `make -n systest` now emits `build/bin/systest_persistence`
- `make -n coverage-ut` now builds `build-cov/bin/systest_persistence`

**3. Informational-only WL/GC assertions**
Replaced all `TEST_ASSERT(true)` placeholders with real predicates tied to each case's stated goal:

| Case | New assertion |
|------|---------------|
| WL-001 | `avg_erase > 0` (blocks actually erased) + `max >= avg >= min` ordering |
| WL-002 | workload produced erase skew exceeding threshold + `wear_level_should_run_static()` returns true |
| WL-005 | blocks erased even with dynamic WL disabled AND `move_count == 0` (dynamic WL didn't run) |
| GC-003 | GC actually ran during overwrite phase AND post-GC latency within 100x baseline |

Total assertions now **48/48 PASS** in systest_wear_gc (up from 44/44 placeholders). All other systest binaries remain green.

PR #68, opened 2026-04-10
fix: address 4 Codex findings + CB-004/007 + YAML parser fix

AlvinYangZF:
## Summary

Fixes 4 real bugs flagged in a Codex review of current master, validated in an isolated worktree against commit `71318ca`.

| # | Bug | Where | Severity |
|---|-----|-------|----------|
| 1 | `lat_monitor_percentile()` hits \`1ULL << 64\` (undefined behavior) on the highest histogram bucket | `src/controller/latency_monitor.c:90,95` | Correctness (UB) |
| 2 | `media_create_checkpoint()` / `media_restore_checkpoint()` silently discard the caller-supplied \`checkpoint_dir\` and write a hardcoded \`checkpoint.bin\` in the cwd | `src/media/media.c:939,949,959` | API contract + cross-run contamination |
| 3 | `wear_level_set_alert_threshold()` declared in the public header but has no definition — callers hit an undefined-symbol link error | `include/ftl/wear_level.h:49` | Missing implementation |
| 4 | `systest_persistence` PR-001 / PR-003 print \`(verified 0 of 500 LBAs)\` and still pass because the assertion is only \`rc != HFSSS_ERR_IO\` | `tests/systest_persistence.c:234,456` | Test quality (false coverage) |

### 1. Latency histogram overflow
Clamp the shift to `QOS_HIST_BUCKETS - 1` and return `UINT64_MAX` for the top bucket and the no-target-hit fallthrough. Verified with a standalone repro that drops one sample into bucket 63 — now returns `UINT64_MAX` instead of `1`.

### 2. Checkpoint directory handling
Added `build_ckpt_path()` helper that joins the caller's directory with \`checkpoint.bin\` / \`checkpoint_inc.bin\`, `mkdir(0755)`s the directory when missing, and validates against path truncation. Confirmed with \`systest_persistence\` — no more \`checkpoint.bin\` left in the repo root after the run.

### 3. `wear_level_set_alert_threshold` implementation
Added the definition in `src/ftl/wear_level.c` with sanity checks:
- `alert_threshold <= 100`
- `critical_threshold <= 100`
- `critical_threshold >= alert_threshold` (keeps the alert hierarchy well-formed)

Verified with a standalone link repro — previously \`Undefined symbols\`, now links and runs.

### 4. PR-001 / PR-003 real round-trip assertion
The original design claimed a full save/restore round-trip but could never succeed because \`media_load()\` restores NAND bytes while the fresh FTL has an empty L2P table — so reads through \`nvme_uspace_read\` returned NOENT for every LBA. The assertion \`rc != HFSSS_ERR_IO\` was trivially true with 0 verified.

Fix: snapshot the L2P mapping before teardown via \`mapping_l2p()\`, then restore it into the reloaded device via \`mapping_direct_set()\` after \`media_load()\` / \`media_restore_checkpoint()\`. This closes the full media + FTL round-trip and the assertion now checks \`verified == write_count\`. Before: \`verified 0 of 500\`. After: \`verified 500 of 500\`.

## Test plan
- [x] systest_persistence: 39/39 PASS, verified 500/500 LBAs
- [x] systest_data_integrity: 50/50 PASS
- [x] systest_nvme_compliance: 70/70 PASS
- [x] systest_error_boundary: 536/536 PASS
- [x] systest_performance: 23/23 PASS
- [x] No checkpoint.bin left in cwd
- [x] Standalone repros confirm bugs 1 and 3
- [ ] CI validation

### DIFF
--- a/src/common/hfsss_config.c
+++ b/src/common/hfsss_config.c
@@ -4,6 +4,7 @@
 #include "common/common.h"
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <string.h>
 #include <ctype.h>
 #include <errno.h>
@@ -159,6 +160,10 @@ int hfsss_config_load(const char *path, struct hfsss_config *cfg,
 
     while (fgets(line, sizeof(line), f)) {
         lineno++;
+        /* Detect whether the line is indented BEFORE trimming, so we can
+         * tell section children (indented) from top-level keys (flush-left).
+         * A top-level non-blank line resets the current section. */
+        bool indented = (line[0] == ' ' || line[0] == '\t');
         char *p = trim(line);
         if (*p == '#' || *p == '\0') continue;
 
@@ -177,6 +182,13 @@ int hfsss_config_load(const char *path, struct hfsss_config *cfg,
         }
 
         unquote(val);
+
+        /* Top-level key (no indentation) closes any active section so that
+         * e.g. "log_level: 3" after a "persist:" block is applied at the
+         * root, not as "persist.log_level". */
+        if (!indented) {
+            section[0] = '\0';
+        }
 
         if (apply_kv(cfg, section, key, val) != 0) {
             HFSSS_LOG_WARN("CFG", "unknown key '%s.%s' at line %d (ignored)",
--- a/src/controller/latency_monitor.c
+++ b/src/controller/latency_monitor.c
@@ -1,4 +1,5 @@
 #include "controller/qos.h"
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -86,13 +87,18 @@ u64 lat_monitor_percentile(const struct ns_latency_monitor *mon,
     for (u32 i = 0; i < QOS_HIST_BUCKETS; i++) {
         cumulative += mon->buckets[i];
         if (cumulative >= target) {
-            /* Return upper bound of this bucket in microseconds */
+            /* Return upper bound of this bucket in microseconds.
+             * Clamp the shift: at i == QOS_HIST_BUCKETS - 1 (last bucket)
+             * a shift by QOS_HIST_BUCKETS would be UB on a 64-bit int. */
+            if (i + 1 >= QOS_HIST_BUCKETS) {
+                return UINT64_MAX;
+            }
             return (1ULL << (i + 1));
         }
     }
 
-    /* Should not reach here; return max bucket upper bound */
-    return (1ULL << QOS_HIST_BUCKETS);
+    /* Exhausted all buckets without reaching target: return saturated max */
+    return UINT64_MAX;
 }
 
 bool lat_monitor_check_sla(struct ns_latency_monitor *mon)
--- a/src/ftl/wear_level.c
+++ b/src/ftl/wear_level.c
@@ -47,6 +47,24 @@ int wear_level_set_static_threshold(struct wear_level_ctx *ctx, u32 threshold)
     return HFSSS_OK;
 }
 
+int wear_level_set_alert_threshold(struct wear_level_ctx *ctx,
+                                   u32 alert_threshold,
+                                   u32 critical_threshold)
+{
+    if (!ctx) {
+        return HFSSS_ERR_INVAL;
+    }
+    /* Sanity: thresholds are percentages (0..100), and critical must be
+     * >= alert for the alert hierarchy to make sense. */
+    if (alert_threshold > 100 || critical_threshold > 100 ||
+        critical_threshold < alert_threshold) {
+        return HFSSS_ERR_INVAL;
+    }
+    ctx->wear_alert_threshold = alert_threshold;
+    ctx->wear_critical_threshold = critical_threshold;
+    return HFSSS_OK;
+}
+
 bool wear_level_should_run_static(struct wear_level_ctx *ctx, u64 current_ts, u32 max_erase, u32 min_erase)
 {
     if (!ctx || !ctx->enabled || !ctx->static_enabled) {
--- a/src/media/media.c
+++ b/src/media/media.c
@@ -1,6 +1,10 @@
 #include "media/media.h"
+#include "sssim.h"
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 
 /* Forward declarations */
 static void media_wait_until_available(struct media_ctx *ctx, enum op_type op,
@@ -936,28 +940,70 @@ int media_load(struct media_ctx *ctx, const char *filepath)
     return HFSSS_OK;
 }
 
-int media_create_checkpoint(struct media_ctx *ctx, const char *checkpoint_dir)
+/* Join a directory and filename into out buffer. Caller must ensure out_sz
+ * is large enough; returns HFSSS_ERR_INVAL on truncation. */
+static int build_ckpt_path(char *out, size_t out_sz, const char *dir,
+                           const char *filename)
 {
-    (void)checkpoint_dir; /* TODO: Implement directory-based checkpointing */
-    int ret = media_save(ctx, "checkpoint.bin");
-    if (ret == HFSSS_OK) {
-        media_mark_all_clean(ctx);
+    if (!out || !dir || !filename || out_sz == 0) {
+        return HFSSS_ERR_INVAL;
     }
-    return ret;
+    int n = snprintf(out, out_sz, "%s/%s", dir, filename);
+    if (n < 0 || (size_t)n >= out_sz) {
+        return HFSSS_ERR_INVAL;
+    }
+    return HFSSS_OK;
 }
 
-int media_create_incremental_checkpoint(struct media_ctx *ctx, const char *checkpoint_dir)
+int media_create_checkpoint(struct media_ctx *ctx, const char *checkpoint_dir)
 {
-    (void)checkpoint_dir; /* TODO: Implement directory-based checkpointing */
-    int ret = media_save_incremental(ctx, "checkpoint_inc.bin");
-    if (ret == HFSSS_OK) {
+    if (!ctx || !checkpoint_dir) {
+        return HFSSS_ERR_INVAL;
+    }
+    /* Ensure directory exists (idempotent) */
+    mkdir(checkpoint_dir, 0755);
+
+    char path[SSSIM_PATH_LEN];
+    int rc = build_ckpt_path(path, sizeof(path), checkpoint_dir,
+                             "checkpoint.bin");
+    if (rc != HFSSS_OK) return rc;
+
+    rc = media_save(ctx, path);
+    if (rc == HFSSS_OK) {
         media_mark_all_clean(ctx);
     }
-    return ret;
+    return rc;
+}
+
+int media_create_incremental_checkpoint(struct media_ctx *ctx,
+                                        const char *checkpoint_dir)
+{
+    if (!ctx || !checkpoint_dir) {
+        return HFSSS_ERR_INVAL;
+    }
+    mkdir(checkpoint_dir, 0755);
+
+    char path[SSSIM_PATH_LEN];
+    int rc = build_ckpt_path(path, sizeof(path), checkpoint_dir,
+                             "checkpoint_inc.bin");
+    if (rc != HFSSS_OK) return rc;
+
+    rc = media_save_incremental(ctx, path);
+    if (rc == HFSSS_OK) {
+        media_mark_all_clean(ctx);
+    }
+    return rc;
 }
 
 int media_restore_checkpoint(struct media_ctx *ctx, const char *checkpoint_dir)
 {
-    (void)checkpoint_dir; /* TODO: Implement directory-based checkpointing */
-    return media_load(ctx, "checkpoint.bin");
+    if (!ctx || !checkpoint_dir) {
+        return HFSSS_ERR_INVAL;
+    }
+    char path[SSSIM_PATH_LEN];
+    int rc = build_ckpt_path(path, sizeof(path), checkpoint_dir,
+                             "checkpoint.bin");
+    if (rc != HFSSS_OK) return rc;
+
+    return media_load(ctx, path);
 }
--- a/tests/systest_error_boundary.c
+++ b/tests/systest_error_boundary.c
@@ -8,6 +8,11 @@
 #include "pcie/nvme_uspace.h"
 #include "ftl/ftl.h"
 #include "common/common.h"
+#include "common/hfsss_config.h"
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/stat.h>
 
 /* --------------- Test Harness --------------- */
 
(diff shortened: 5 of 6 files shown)


AlvinYangZF:
Posting the consolidated review log here for tracking.

Round 2 items are the 4 findings already summarized in this PR body and addressed by this change:
- `lat_monitor_percentile()` highest-bucket overflow / UB in `src/controller/latency_monitor.c`
- checkpoint APIs ignoring caller-supplied `checkpoint_dir` in `src/media/media.c`
- missing public implementation of `wear_level_set_alert_threshold()`
- `systest_persistence` PR-001 / PR-003 passing with `verified 0 of 500` because the assertion only checked `rc != HFSSS_ERR_IO`

Additional follow-up findings from the next review rounds against current `origin/master` (`71318ca`):

Round 3:
1. `ftl_mt_start()` can deadlock on partial worker-thread startup failure. If `pthread_create()` fails after some workers have started, the failure path pushes `IO_OP_STOP` and immediately `pthread_join()`s already-started workers, but never signals `request_cond`. Local wrapped-thread repro timed out instead of returning.
2. `nbd_async_start()` can deadlock if CQ thread creation fails after SQ starts. The failure path joins `sq_thread` without first shutting down the socket to unblock the SQ thread's blocking `recv()`. Local wrapped-thread repro timed out.
3. `bench_run()` reports success even when worker thread creation fails. It ignores `pthread_create()` / `pthread_join()` return values and can return `HFSSS_OK` with inconsistent results. Local repro: `rc=0 total_ops=9 hist_sum=3`.
4. `bench_run()` silently drops requested operations when `op_count` is not divisible by `num_threads`. Local repro: `op_count=10, num_threads=3` returned `total_ops=9`.

Round 4:
1. `mem_pool_alloc()` exposes allocator metadata as user memory. Overwriting an allocated block can destroy the in-band `mem_block` header and permanently leak the block. Local repro: allocate one block, `memset()` it to zero, free it, then the next allocation returns `NULL` and the pool still reports `used=1 free=0`.
2. `trace_ring_write()` corrupts ring accounting on overflow. It increments `write_idx` before checking fullness, so dropped entries still become visible to `trace_ring_pending()` / `trace_ring_read()`. Local repro with `HFSSS_TRACE_TEST_MODE`: a 256-slot ring reported `pending=257` and returned 257 readable entries after 257 writes with one `NOSPC`.
3. `watchdog_feed()` cannot recover a timed-out task even though the function tries to. It rejects every non-`WATCHDOG_TASK_ACTIVE` task up front, so the later `WATCHDOG_TASK_TIMEOUT -> ACTIVE` transition is unreachable. Local repro: feeding a timed-out task returned `HFSSS_ERR_NOENT` and left it in `WATCHDOG_TASK_TIMEOUT`.
4. `ftl_mt_start()` treats GC/WL background-thread startup failure as success. It checks worker-thread creation but ignores `pthread_create()` return values for the GC and WL threads. Local wrapped-thread repro forcing calls 9 and 10 to fail still returned `rc=0` with `gc_running=1 wl_running=1`.

Round 5:
1. `bench_run()` reports completely wrong percentiles for very small samples. `hist_percentile()` truncates `total * pct / 100.0` to zero for `total == 1`, so a one-sample run returns bucket 0 regardless of where the only sample actually landed. Local repro: `p50=1 p99=1 p999=1` while the only histogram sample was in bucket 6.
2. `log_init()` accepts `buffer_size == 0` and creates a context that later hits undefined behavior in `log_printf()`. Local init-only repro returned `rc=0 buffer_size=0`.
3. `mutex_lock()` exposes a `timeout_ns` parameter but ignores it completely and always blocks in `pthread_mutex_lock()`. Local repro: `mutex_lock(..., 50ms)` waited about `206.9ms` until the other thread released the lock, then returned success.
4. `hfsss_nbd_server` does not shut down promptly on `SIGINT` / `SIGTERM` if a client session is active. The signal handler only clears `g_running` and closes the listening socket; the active-session path stays blocked in `pthread_join(async_ctx.sq_thread)` / `pthread_join(async_ctx.cq_thread)` without closing the active `client_fd` or calling `nbd_async_stop()`. This one is from code-path inspection rather than a runtime repro.

I have the full local notes in `review.md` as Round 2 through Round 5, and I am continuing the next review round from there.

AlvinYangZF:
Added follow-up commit \`f9623cf\` that expands scope:

**New test cases (Plan 2 quick wins)**
- **CB-004**: High OP ratio (50%) — 80% fill, all writes/overwrites succeed, low GC pressure, WAF >= 1.0
- **CB-007**: Config file round-trip — hfsss_config save/load, asserts 7 fields round-trip and both versions validate

**Latent bug discovered by CB-007**: `hfsss_config_load()` was dropping top-level keys that appeared after a section. Root cause: leading whitespace was stripped before the section-scope check, so a flush-left \`log_level: 3\` after a \`persist:\` block was interpreted as \`persist.log_level\` and dropped as unknown.

**Fix**: detect indentation on the original line (before trim). Flush-left non-blank lines now reset the section scope, restoring correct YAML-ish behavior.

**Results**: systest_error_boundary now **556/556 PASS** (was 555/556 with CB-007 \`log_level round-trips\` failing).

System Test Plan Phase 2 progress: **73 → 75 / 107 cases** (+2%).

AlvinYangZF:
Additional master review findings from later passes against `origin/master` at `71318cac8f6153719d9df580115729a6425214a2` (`2026-04-10`). These are the Round 6 / 7 / 8 items from my local review log.

## Round 6

1. `proc_interface_init()` fails with its own default path unless the parent directory already exists.
   - `src/common/proc_interface.c` defaults to `/tmp/hfsss/proc` but only does a single `mkdir(proc->dir, 0755)`.
   - If `/tmp/hfsss` is absent, init fails with `ENOENT`.
   - Local repro: both `proc_interface_init(&proc, NULL)` and a custom nested path failed when the parent directory was missing.

2. `ns_token_bucket_refill()` permanently loses sub-token elapsed time on frequent calls.
   - `src/common/reliability.c` computes whole tokens from elapsed time, then always advances `last_refill_ns` to `now_ns`.
   - Any fractional remainder smaller than one token is discarded instead of being carried forward.
   - Local repro: with `refill_rate = 1 token/sec`, starting from zero tokens and calling refill ten times 100ms apart still left the bucket at `0` tokens after one full second.

3. `meta_mirror_verify()` does not verify payload integrity at all.
   - `src/common/reliability.c` compares only the two headers' `crc32` and `payload_len`.
   - It never reads either payload and never recomputes CRCs.
   - Local repro: after `meta_mirror_write()`, corrupting one byte in the primary payload still left `meta_mirror_verify()` returning `HFSSS_OK`.

4. `perf_validation_report_to_json()` can emit invalid JSON while returning success.
   - `src/perf/perf_validation.c` interpolates `req_id`, `description`, and `unit` with raw `%s` formatting and performs no JSON escaping.
   - Local repro: a report containing quotes and backslashes serialized to invalid JSON and still returned `rc=0`.

## Round 7

1. `thermal_compute_level()` can drop multiple thermal levels in a single call even though the code comment promises one-level step-down behavior.
   - `src/common/thermal.c` uses a sequence of mutating `if` checks on `current_level`.
   - Local repro: `thermal_compute_level(THERMAL_LEVEL_SHUTDOWN, 60)` returned `THERMAL_LEVEL_NONE` instead of stepping down by one level.

2. `resource_free()` accepts arbitrary pointers and can poison the resource free list.
   - `src/controller/resource.c` validates only `mgr`, `type`, and non-null `ptr`; it never checks ownership/range.
   - Local repro: after one valid allocation, freeing a stack address caused the next `resource_alloc()` to return that fake pointer.

3. `ns_mapping_create()` can assign a namespace beyond `total_lbas` once holes appear.
   - `src/ftl/ns_mapping.c` checks capacity with `allocated_lbas + lba_count <= total_lbas`, but placement uses the highest active namespace end rather than reusing holes.
   - Local repro: on a 10-LBA mapping, creating `5 + 5`, deleting the first namespace, then creating another `5` produced a namespace at `lba_start=10, lba_count=5`.

4. `integrity_test_run()` ignores `capacity_bytes` entirely.
   - `integrity_test_cfg` exposes `capacity_bytes`, but `src/common/reliability.c` never reads it.
   - Local repro: with `capacity_bytes=4096`, `block_size=4096`, and `num_blocks=2`, `integrity_test_run()` still returned success with `blocks_written=2` and `verified_blocks=2`.

## Round 8

1. The controller-side shared-memory protocol does not match the kernel-side protocol, so the two halves cannot interoperate.
   - `src/controller/shmem_if.c` / `include/controller/shmem_if.h` model a single ring of `cmd + ready/done`.
   - `src/kernel/hfsss_nvme_shmem.c` implements a different ABI: submit ring + completion ring with separate head/tail indices.
   - `shmem_if_send_cpl()` never copies completion payload anywhere because the controller-side layout has no completion slot at all; it only sets `done = 1`.

2. `arbiter_free_cmd()` corrupts queue accounting when freeing the only command in a queue.
   - `src/controller/arbiter.c` removes a command from its queue only when `cmd->prev || cmd->next` is true.
   - Local repro: enqueuing one command and then freeing it left `count=1` and a non-NULL queue head while `total_cmds` had already dropped to `0`.

3. `rc_lookup()` can return a false cache hit before anything has ever been inserted.
   - `src/controller/read_cache.c` walks the full LRU list without checking whether an entry is actually populated.
   - Local repro: immediately after `rc_init()`, `rc_lookup(&ctx, 0, 16, out)` returned `HFSSS_OK`, incremented `hit_count`, and left `used_count == 0`.

4. `wb_flush()` permanently strands flushed entries instead of recycling them.
   - `src/controller/write_buffer.c` marks each dirty entry `WB_FLUSHED`, clears `dirty_list`, and resets `dirty_count`, but never returns those entries to `free_list`.
   - Local repro: with a one-entry buffer, one successful `wb_write()` + `wb_flush()` left the next `wb_write()` failing with `HFSSS_ERR_NOMEM`.

5. `SCHED_DWRR` enqueue works, but dequeue is effectively unimplemented.
   - `src/controller/scheduler.c` has a `SCHED_DWRR` branch in `scheduler_enqueue()` but `scheduler_dequeue()` handles only FIFO/Greedy/Deadline/WRR.
   - Local repro: with an initialized DWRR scheduler and one pending command for `nsid=7`, enqueue succeeded, `pending_cmds` became `1`, and `scheduler_dequeue()` still returned `NULL`.

I have local repros for each of the above if it would help to turn any of these into more focused follow-up fixes.

AlvinYangZF:
One follow-up on the new `systest_persistence` assertions:

`PR-001` / `PR-003` are definitely stronger now, but they still don't fully prove that the restored mapping state is complete. In `tests/systest_persistence.c`, the new restore path repopulates entries with `mapping_direct_set()` before read-back. The problem is that `mapping_direct_set()` explicitly restores only L2P; it does **not** rebuild P2L or `valid_count` (see `src/ftl/mapping.c`, comment and implementation around `mapping_direct_set()` / `mapping_rebuild_p2l()`). The real recovery flow in `src/ftl/superblock.c` follows direct L2P restore with `mapping_rebuild_p2l()` for exactly that reason.

I reproed this locally against the PR head with a minimal mapping-only test:
- after `mapping_direct_set()`, `valid_count` was still `0`
- `mapping_p2l()` still returned `HFSSS_ERR_NOENT`
- only after `mapping_rebuild_p2l()` did the mapping state become internally consistent

So the new assertions do prove "reads work after manually restoring L2P", but they still don't prove "the recovered FTL mapping state is fully restored and internally consistent". I think the test should either:
1. call `mapping_rebuild_p2l()` after the snapshot restore, to match the real recovery contract, or
2. narrow the test wording/comments so it doesn't claim a full recovered mapping state when only L2P has been restored.

AlvinYangZF:
One more follow-up on the checkpoint-directory fix:

`media_create_checkpoint()` / `media_create_incremental_checkpoint()` now respect `checkpoint_dir`, but they still don't actually guarantee that the directory exists for nested paths. The new code only does a single `mkdir(checkpoint_dir, 0755)` and ignores the return value before calling `media_save()` / `media_save_incremental()`.

That works if the parent already exists, but it still fails for a missing nested path like `/tmp/foo/bar/baz` because `mkdir()` does not create parents recursively. I reproed this locally against the PR head with a minimal media-only test:

- `media_create_checkpoint(&ctx, "/tmp/pr68-missing-parent/a/b")`
- returned `HFSSS_ERR_IO` (`-100`)

So this fixes the original "hardcoded cwd filename" bug, but not the broader "arbitrary caller-supplied checkpoint_dir works" expectation. If you want this API to accept nested paths, it probably needs recursive parent creation (or an explicit validation/error path that surfaces the real mkdir failure earlier instead of falling through to `media_save()`).

AlvinYangZF:
Follow-up on the persistence test changes: `PR-001` / `PR-003` still restore only L2P, not a fully consistent mapping state.

In `tests/systest_persistence.c`, both restore paths rebuild visibility by calling `mapping_direct_set()` for each saved LBA. But `mapping_direct_set()` is explicitly documented in `src/ftl/mapping.c` as *not* rebuilding P2L or `valid_count`; the real recovery flow in `src/ftl/superblock.c` immediately follows it with `mapping_rebuild_p2l()`.

I re-ran a targeted repro against this PR head and still get:
- `after_direct_set rc=0 valid_count=0`
- `p2l_lookup_rc=-6`
- `after_rebuild rc=0 valid_count=1`

So these assertions now prove "reads work after manually restoring L2P", but they still do not prove that the recovered FTL mapping state is complete and internally consistent the way the production restore path requires.

AlvinYangZF:
The checkpoint path fix still only handles a single-level directory.

`media_create_checkpoint()` / `media_create_incremental_checkpoint()` now join `checkpoint_dir` with the filename, which fixes the old cwd issue, but they still just do a plain `mkdir(checkpoint_dir, 0755)` and ignore the return value. If the caller passes a nested path whose parent does not exist, the save still fails.

I re-ran the same repro against this PR head with `"/tmp/pr68-missing-parent/a/b"` and `media_create_checkpoint()` still returns `-100`.

So this fixes "respect the caller-supplied directory", but not yet "ensure an arbitrary checkpoint_dir exists before writing into it".

AlvinYangZF:
Deep rereview on merged PR #68 at `f9623cf` found three remaining issues worth recording.

1. `tests/systest_persistence.c` PR-001 / PR-003 still do not restore a consistent mapping state. The reload path repopulates only L2P via `mapping_direct_set()` (`tests/systest_persistence.c`), but `mapping_direct_set()` is explicitly documented in `src/ftl/mapping.c` as *not* rebuilding P2L or `valid_count`; that is what `mapping_rebuild_p2l()` is for. Because the tests then only verify readback, they can pass while leaving derived mapping state inconsistent, so they still overclaim a full post-recovery FTL state.

2. The checkpoint-directory fix in `src/media/media.c` is still incomplete for nested paths. `media_create_checkpoint()` and `media_create_incremental_checkpoint()` do a single `mkdir(checkpoint_dir, 0755)` and ignore the result. That handles an already-existing flat directory, but it still fails for caller-supplied nested paths when parents do not exist. So the hardcoded filename bug is fixed, but the directory-based API is still not robust.

3. `CB-004` in `tests/systest_error_boundary.c` does not actually validate the behavior it claims. The comments say 50% OP should keep GC pressure light and WAF low, but the only assertion is `stats.waf >= 1.0`, which is just a sanity check that stats are not broken. A major WAF regression at 50% OP would still pass.

This is a static rereview of the merged PR head; I did not rerun the branch locally in this pass.

PR #69, opened 2026-04-10
fix: resolve 4 concurrency deadlocks (Codex Round 3)

AlvinYangZF:
## Summary

Fixes 4 pthread-related concurrency bugs flagged by Codex review (Round 3), all reproduced locally with wrapped-thread repros against current master.

Independent of PR #68 (which covers Round 2 findings + CB-004/007). This PR touches a disjoint set of files (\`ftl_worker.c\`, \`nbd_async.c\`, \`perf_validation.c\`).

### 1. \`ftl_mt_start()\` partial-startup deadlock
If \`pthread_create()\` fails for worker N after N-1 workers have started, the cleanup loop pushed \`IO_OP_STOP\` into each request ring and immediately called \`pthread_join()\` — but **never signaled \`request_cond\`**. Workers sleeping in \`pthread_cond_wait()\` never woke to see the stop request, so the join deadlocked.

**Fix**: Extracted a \`rollback_started_workers()\` helper that mirrors \`ftl_mt_stop()\`: push STOP, acquire \`request_lock\`, signal \`request_cond\`, release, then join.

**Bonus**: Also fixed the Round 4 companion bug where \`pthread_create()\` for the GC and WL background threads in \`ftl_mt_start()\` was unchecked. If either fails, we now unwind the previous bring-up steps cleanly: tear down GC (if it started), destroy the gc cond/mutex, and rewind all already-started workers.

### 2. \`nbd_async_start()\` CQ-failure deadlock
When CQ thread creation fails after SQ has started, the cleanup path called \`pthread_join(sq_thread)\` without first shutting down the client socket. The SQ thread stayed blocked in \`recv()\` and the join never returned.

**Fix**: Call \`shutdown(client_fd, SHUT_RD)\` before joining, matching the pattern used in \`nbd_async_stop()\`.

### 3. \`bench_run()\` ignored pthread return codes
Failing worker creation could leave \`bench_run()\` reporting \`rc=HFSSS_OK\` with inconsistent aggregated state.

**Fix**: Track per-worker \`started[]\` state, unwind on create failure (zero the \`op_count\` budget so in-flight workers exit quickly, then join), and return \`HFSSS_ERR_INVAL\` / \`HFSSS_ERR_IO\` on create / join failure respectively.

### 4. \`bench_run()\` silently dropped operations
A request of \`op_count=10, num_threads=3\` returned \`total_ops=9\` because each worker got \`(op_count / nt) = 3\` from integer division.

**Fix**: Distribute the remainder across the first \`(op_count % nt)\` workers so total ops match exactly.

## Test plan
- [x] \`test_mt_ftl\`: 10/10 PASS
- [x] \`test_perf_validation\`: 62/62 PASS
- [x] \`systest_data_integrity\`: 50/50 PASS
- [x] \`systest_nvme_compliance\`: 70/70 PASS
- [x] \`systest_error_boundary\`: 536/536 PASS (pre-PR-68 baseline)
- [x] \`systest_performance\`: 23/23 PASS
- [x] \`systest_persistence\`: 39/39 PASS
- [x] \`systest_wear_gc\`: 48/48 PASS
- [x] Standalone repro for bench_run divisibility returns \`total_ops=10\` for \`op_count=10, num_threads=3\` (was 9 before)
- [ ] CI validation

### DIFF
--- a/src/ftl/ftl_worker.c
+++ b/src/ftl/ftl_worker.c
@@ -11,7 +11,7 @@ extern void *wl_thread_main(void *arg);
 
 /* Global pointers set by ftl_mt_start() so worker threads can signal GC */
 pthread_mutex_t *ftl_mt_gc_mutex_ptr = NULL;
-pthread_cond_t  *ftl_mt_gc_cond_ptr  = NULL;
+pthread_cond_t *ftl_mt_gc_cond_ptr = NULL;
 
 static void *ftl_worker_main(void *arg)
 {
@@ -43,9 +43,9 @@ static void *ftl_worker_main(void *arg)
             u32 data_step = req.data_stride ? req.data_stride : page_size;
             u8 *ptr = req.data;
             for (u32 i = 0; i < req.count; i++) {
-                rc = ftl_read_page_mt(w->ftl, w->taa,
-                                      req.lba + (u64)i * lba_step, ptr);
-                if (rc != HFSSS_OK) break;
+                rc = ftl_read_page_mt(w->ftl, w->taa, req.lba + (u64)i * lba_step, ptr);
+                if (rc != HFSSS_OK)
+                    break;
                 ptr += data_step;
             }
             break;
@@ -56,9 +56,9 @@ static void *ftl_worker_main(void *arg)
             u32 data_step = req.data_stride ? req.data_stride : page_size;
             const u8 *ptr = req.data;
             for (u32 i = 0; i < req.count; i++) {
-                rc = ftl_write_page_mt(w->ftl, w->taa,
-                                       req.lba + (u64)i * lba_step, ptr);
-                if (rc != HFSSS_OK) break;
+                rc = ftl_write_page_mt(w->ftl, w->taa, req.lba + (u64)i * lba_step, ptr);
+                if (rc != HFSSS_OK)
+                    break;
                 ptr += data_step;
             }
             break;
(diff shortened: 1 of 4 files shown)


AlvinYangZF:
`bench_run()` still does not actually stop already-started workers promptly on `pthread_create()` failure.

The rollback path in `src/perf/perf_validation.c` now zeroes `ws[i].op_count`, but the worker copies that value into local `max_ops` once at startup and never rereads it:
- worker snapshot: `max_ops = ws->op_count`
- rollback path later mutates `ws[i].op_count = 0`

I verified this with a targeted repro that forces the second `pthread_create()` to fail while running a `duration_s=1` config. `bench_run()` returned the error, but only after waiting the full second:
- `rc=-2 create_calls=2 elapsed_ms=1000 total_ops=0`

So the deadlock is gone, but this cleanup path still does not do what the comment says for already-started workers. A shared stop flag (or some other state the worker actually consults after startup) is still needed if the goal is to fail fast instead of waiting out the duration.

AlvinYangZF:
Addressed in b2e7c23.

**Fix**: replaced the dead `op_count` zeroing with a shared cancellation flag (`volatile bool stop_flag` on the bench_run stack). Workers consult it via `__atomic_load_n(..., __ATOMIC_ACQUIRE)` at the top of the hot loop; rollback raises it via `__atomic_store_n(..., __ATOMIC_RELEASE)` before joining. Lifetime is the bench_run frame, which extends past every join.

**Test**: added a test seam (`g_pthread_create` function pointer + `__perf_test_set_pthread_create` extern setter, intentionally not in the public header) so the test can inject a failing pthread_create deterministically. Two new assertions in `test_perf_validation.c::test_bench_run_create_fail_fast`:

- `duration_s=5, num_threads=4`, fail on 2nd create → asserts `elapsed_ms < 1000` (would have been ~5000 before the fix)
- `op_count=1M, num_threads=2`, fail on 1st create → asserts `elapsed_ms < 500`

**Verification**:
- `test_perf_validation`: 67/67 PASS (5 new asserts)
- `test_common`: 132/132 PASS
- `systest_performance`: 23/23 PASS

PR #70, opened 2026-04-11
fix: resolve 7 bugs from Codex review rounds 4 & 5

AlvinYangZF:
## Summary

Fixes 7 bugs flagged by Codex review rounds 4 and 5. All reproduced locally with standalone repros against current master.

**Depends on #69** (Round 3 concurrency deadlocks) — this PR is based off that branch. Will auto-rebase to master once #69 merges.

## Round 4 (3 bugs)

> Round 4 #4 (\`ftl_mt_start\` unchecked GC/WL pthread_create) was folded into PR #69 since it's in the same file as the Round 3 deadlock fixes.

### R4 #1 — \`mem_pool_alloc()\` exposed allocator metadata as user memory
The old layout put \`struct mem_block\` at the start of every user block. Any caller writing into the allocated block destroyed the in-band \`next\`/\`in_use\` header and permanently leaked the block. Standalone repro: alloc → memset(0) → free → next alloc returned NULL with \`free=0\`.

**Fix**: moved metadata to a parallel \`pool->blocks[]\` array. User memory at index \`i\` lives at \`pool->memory + i * block_size\`. \`mem_pool_free()\` derives the index from the pointer offset.

### R4 #2 — \`trace_ring_write()\` corrupted accounting on overflow
Unconditional \`__sync_fetch_and_add(write_idx, 1)\` advanced the write index BEFORE checking fullness, so a rejected NOSPC write still bumped \`write_idx\`. A 256-slot ring reported \`pending=257\` after 257 writes with 1 NOSPC.

**Fix**: CAS loop claims the slot only if fullness check passes.

### R4 #3 — \`watchdog_feed()\` made its own recovery unreachable
Early-return at \`state != WATCHDOG_TASK_ACTIVE\` fired before the \`TIMEOUT → ACTIVE\` transition code below, so a timed-out task could never be fed back.

**Fix**: accept feeds for both ACTIVE and TIMEOUT. INACTIVE tasks still need \`watchdog_enable_task()\` first.

### R5 #1 — \`hist_percentile()\` wrong for small samples
For \`total == 1\`, \`(uint64_t)(1 * 50.0 / 100.0)\` truncated \`0.5\` to \`0\`. Then \`cum >= target\` fired at bucket 0 regardless of the sample's real position. A one-sample run always reported \`p50/p99/p999 = bucket 0 = 1us\`.

**Fix**: use \`ceil()\` with minimum 1 and cap by total. Standalone repro verifies a single sample in bucket 6 now reports \`p99 = 64us\`, not \`1\`.

### R5 #2 — \`log_init()\` accepted \`buffer_size == 0\`
Built a context that later hit UB in \`log_printf()\` via \`(head + 1) % 0\`.

**Fix**: reject \`buffer_size == 0\` with \`HFSSS_ERR_INVAL\`.

### R5 #3 — \`mutex_lock(timeout_ns)\` ignored its timeout
\`timeout_ns\` was \`(void)\`-cast. Callers asking for a 50ms budget waited 200ms+. Standalone repro: 50ms timeout returned \`HFSSS_ERR_BUSY\` in 50ms (was 200ms+).

**Fix**: when \`timeout_ns > 0\`, poll \`pthread_mutex_trylock\` in a 100us-sleep loop with a CLOCK_MONOTONIC deadline. \`timeout_ns == 0\` preserves the original block-forever semantics. \`pthread_mutex_timedlock\` is not portable to macOS, and the mutex is recursive, so we emulate.

### R5 #4 — \`hfsss-nbd-server\` didn't respond to SIGINT during active session
Signal handler only cleared \`g_running\` and closed the listening fd. Main thread was stuck in \`pthread_join(sq_thread)\` / \`pthread_join(cq_thread)\`, with both threads blocked in \`recv()\` on the active \`client_fd\`.

**Fix**: publish the active \`client_fd\` as a volatile global set before serving a session and cleared after. Signal handler \`shutdown()\`s it, which unblocks \`recv()\` / \`read()\` in the serving threads. \`shutdown()\` is async-signal-safe.

## Test plan

- [x] \`test_common\` 132/132 PASS (covers mempool)
- [x] \`test_perf_validation\` 62/62 PASS (covers hist_percentile)
- [x] \`systest_nvme_compliance\` 70/70 PASS
- [x] \`systest_persistence\` 39/39 PASS
- [x] \`systest_error_boundary\` 536/536 PASS
- [x] \`systest_performance\` 23/23 PASS
- [x] R4#1 repro: alloc + memset + free + realloc works, \`free=3\` (was NULL + \`free=0\`)
- [x] R4#3 repro: \`watchdog_feed(TIMEOUT)\` returns \`rc=0\` and transitions to ACTIVE
- [x] R5#1 repro: single sample in bucket 6 reports \`p99=64us\` (was 1)
- [x] R5#3 repro: 50ms timeout returns BUSY in 50ms (was 200ms+)
- [ ] systest_data_integrity / systest_wear_gc (running in background)
- [ ] CI validation

### DIFF
--- a/include/common/mempool.h
+++ b/include/common/mempool.h
@@ -7,7 +7,9 @@
 extern "C" {
 #endif
 
-/* Memory Pool Block */
+/* Memory Pool Block Metadata.
+ * Kept in a parallel array, NOT embedded in user memory, so that
+ * user writes to allocated blocks cannot corrupt allocator state. */
 struct mem_block {
     struct mem_block *next;
     int in_use;
@@ -19,7 +21,8 @@ struct mem_pool {
     u32 block_count;
     u32 free_count;
     u32 used_count;
-    void *memory;
+    void *memory;               /* user data area: block_count * block_size */
+    struct mem_block *blocks;   /* metadata array: block_count entries */
     struct mem_block *free_list;
     void *lock;
     u64 alloc_count;
--- a/include/common/mempool.h
+++ b/include/common/mempool.h
@@ -15,13 +15,25 @@ struct mem_block {
     int in_use;
 };
 
+/* Minimum alignment for user-data pointers returned by mem_pool_alloc.
+ * Matches the effective guarantee glibc malloc() provides for any
+ * standard type (max_align_t is 16 bytes on x86-64 and aarch64). This
+ * also preserves the historical property of the old implementation,
+ * which spaced slots by sizeof(struct mem_block) == 16 bytes. */
+#define MEMPOOL_MIN_ALIGN 16
+
 /* Memory Pool Context */
 struct mem_pool {
-    u32 block_size;
+    u32 block_size;             /* caller's requested size (for stats) */
+    u32 slot_size;              /* effective stride: block_size rounded
+                                 * up to MEMPOOL_MIN_ALIGN. Used for all
+                                 * address math so that returned pointers
+                                 * stay aligned even when block_size is
+                                 * small (e.g. 1). */
     u32 block_count;
     u32 free_count;
     u32 used_count;
-    void *memory;               /* user data area: block_count * block_size */
+    void *memory;               /* user data area: block_count * slot_size */
     struct mem_block *blocks;   /* metadata array: block_count entries */
     struct mem_block *free_list;
     void *lock;
--- a/include/common/mempool.h
+++ b/include/common/mempool.h
@@ -24,17 +24,17 @@ struct mem_block {
 
 /* Memory Pool Context */
 struct mem_pool {
-    u32 block_size;             /* caller's requested size (for stats) */
-    u32 slot_size;              /* effective stride: block_size rounded
-                                 * up to MEMPOOL_MIN_ALIGN. Used for all
-                                 * address math so that returned pointers
-                                 * stay aligned even when block_size is
-                                 * small (e.g. 1). */
+    u32 block_size; /* caller's requested size (for stats) */
+    u32 slot_size;  /* effective stride: block_size rounded
+                     * up to MEMPOOL_MIN_ALIGN. Used for all
+                     * address math so that returned pointers
+                     * stay aligned even when block_size is
+                     * small (e.g. 1). */
     u32 block_count;
     u32 free_count;
     u32 used_count;
-    void *memory;               /* user data area: block_count * slot_size */
-    struct mem_block *blocks;   /* metadata array: block_count entries */
+    void *memory;             /* user data area: block_count * slot_size */
+    struct mem_block *blocks; /* metadata array: block_count entries */
     struct mem_block *free_list;
     void *lock;
     u64 alloc_count;
--- a/src/common/log.c
+++ b/src/common/log.c
@@ -117,8 +117,7 @@ void log_set_level(struct log_ctx *ctx, u32 level)
     ctx->level = level;
 }
 
-void log_printf(struct log_ctx *ctx, u32 level, const char *module,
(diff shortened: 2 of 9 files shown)


AlvinYangZF:
Final verification complete — all systests pass with the Round 4+5 fixes:

| Test | Result |
|------|--------|
| test_common (mempool) | 132/132 PASS |
| test_perf_validation (hist_percentile) | 62/62 PASS |
| systest_nvme_compliance | 70/70 PASS |
| systest_error_boundary | 536/536 PASS |
| systest_performance | 23/23 PASS |
| systest_persistence | 39/39 PASS |
| systest_data_integrity | **50/50 PASS** (400s) |
| systest_wear_gc | **48/48 PASS** (1456s) |

No regressions from any of the 7 fixes. Standalone repros confirm all behavior changes:
- R4#1 mempool: corrupted user memory no longer breaks allocator
- R4#3 watchdog_feed: TIMEOUT→ACTIVE transition now reachable
- R5#1 hist_percentile: single sample in bucket 6 reports p99=64us (was 1)
- R5#3 mutex_lock: 50ms timeout returns BUSY in 50ms (was 200ms+)

AlvinYangZF:
The mempool fix removes the metadata corruption bug, but it also introduces an alignment regression for small block sizes.

Because blocks are now packed at exactly `block_size` spacing, `mem_pool_alloc()` can return addresses that are no longer naturally aligned when `block_size` is small. I reproed this with `block_size=1` and three allocations; the returned addresses had:
- `mod8: 2 1 0`

So the second and third blocks were not even 2-byte / 4-byte / 8-byte aligned. Any caller that stores wider types in those blocks can now hit UB.

Previously the implementation effectively rounded each slot up to at least `sizeof(struct mem_block)`, so this alignment property existed accidentally. After moving metadata out-of-band, I think the allocator still needs to preserve a minimum alignment for the user-data area instead of spacing blocks by raw `block_size` bytes.

AlvinYangZF:
Addressed in a613941.

**Fix**: added a \`slot_size\` field to \`struct mem_pool\` that is \`block_size\` rounded up to \`MEMPOOL_MIN_ALIGN\` (16, matching glibc \`max_align_t\`). \`mem_pool_alloc()\` and \`mem_pool_free()\` now use \`slot_size\` for all address math; the public \`block_size\` field still reflects the caller's requested value for stats readers. calloc() already returns max_align_t-aligned memory, so base + any multiple of slot_size is also aligned.

**Regression test**: new \`test_mempool_alignment\` in \`tests/test_common.c\` tries block sizes 1, 3, 7, 9, 15, 17 (each non-multiple of 16), allocates all 32 slots per pool, and asserts every returned pointer is 16-byte aligned. Also round-trips each allocation back through \`mem_pool_free\` to confirm the free-path offset math is consistent with the new stride.

With \`block_size=1\` the old code returned addresses with \`mod8: 2 1 0\`. After the fix every returned pointer is \`mod16: 0\`.

**Verification**:
- test_common: 162/162 PASS (30 new alignment assertions)
- test_perf_validation: 67/67 PASS
- test_mt_ftl: 10/10 PASS
- Full \`make all\` clean

**Also**: rebased onto the updated PR #69 tip (which now has the bench_run rollback fix + format cleanup), resolved a trivial conflict in \`hist_percentile\` (kept the ceil() fix, applied clang-format style), and ran clang-format on the 7 files that still had pre-existing violations. PR #70 now has 3 commits on top of #69: R4+5 fix → mempool alignment fix → style cleanup.

PR #71, opened 2026-04-11
[codex] add NAND/media command coverage design

AlvinYangZF:
## Summary

Add a design/spec document for enhancing NAND/Media command coverage with an ONFI/Toggle-standard-first approach.

The spec is focused on the NAND/Media layer only and captures the agreed design direction:

- enterprise TLC/QLC priority
- command coverage first
- ONFI / Toggle standard-equivalent behavior first
- `read parameter page` included in the foundational command set
- `multi-plane` prioritized ahead of `cache read/program`
- vendor-specific behavior deferred behind profile hooks

## What Changed

Added:

- `docs/superpowers/specs/2026-04-10-nand-media-command-coverage-design.md`

The document covers:

- current NAND/Media implementation limits
- standards baseline and public industry context
- proposed command/state/timing architecture
- command coverage tiers
- die/channel/plane execution model
- suspend/resume semantics
- profile model
- phased enhancement roadmap
- validation strategy and risk analysis

## Why

The current NAND simulator models only the basic `read/program/erase` flow with coarse latency and no explicit command engine. That is not sufficient for modeling modern enterprise TLC/QLC NAND behavior such as:

- program/erase suspend-resume
- standard multi-plane operations
- status visibility during in-flight commands
- future standard-equivalent Toggle support

This PR records the agreed design before implementation work starts.

## Validation

- spec reviewed against current `origin/master` code structure
- design decisions updated from discussion and folded into the document
- no code or runtime behavior changed in this PR

PR #72, opened 2026-04-11
[docs] add enterprise controller simulator enhancement design

AlvinYangZF:
## Summary

Add a design document for enhancing the HFSSS controller simulator using public enterprise PCIe Gen5 SSD controller signals.

The design is informed by common capability patterns visible across:

- Marvell Bravera SC5
- Silicon Motion SM8366
- InnoGrit Tacoma IG5669

## What Changed

Added:

- `docs/superpowers/specs/2026-04-11-enterprise-controller-sim-enhancement-design.md`

The document defines:

- current controller-simulator gaps in HFSSS
- a controller-profile-based architecture
- enterprise isolation, mode, media-coupling, and lifecycle layers
- a phased enhancement roadmap
- validation strategy and scope boundaries

## Why

HFSSS already has controller, scheduler, QoS, telemetry, security, and NVMe building blocks, but it does not yet model enterprise controller personas in a coherent way.

This design records a structured path to evolve the controller simulator toward:

- profile-driven enterprise controller behavior
- stronger namespace and QoS isolation
- ZNS / FDP-oriented controller modes
- tighter controller-to-media coupling
- richer telemetry and security lifecycle behavior

## Validation

- reviewed against current `origin/master`
- aligned with existing controller and NAND/media design documents
- no runtime code changed in this PR

### DIFF
--- a/docs/superpowers/specs/2026-04-11-enterprise-controller-sim-enhancement-design.md
+++ b/docs/superpowers/specs/2026-04-11-enterprise-controller-sim-enhancement-design.md
@@ -1,0 +1,496 @@
+# Enterprise Controller Simulator Enhancement Design
+
+**Date:** 2026-04-11  
+**Repo baseline:** `origin/master` at `4155575d1d7e4af8f71db8ae30d74711f63d7fe1`  
+**Scope:** controller simulator and host-facing enterprise behavior  
+**Priority direction:** public enterprise-controller capability first, vendor-private fidelity later  
+**Input sources:** Marvell Bravera SC5 research, Silicon Motion SM8366 research, InnoGrit Tacoma IG5669 research
+
+---
+
+## 1. Summary
+
+This design upgrades HFSSS from a basic single-controller NVMe simulation into a configurable **enterprise SSD controller simulator** informed by three public PCIe Gen5 controller families:
+
+- Marvell **Bravera SC5** (`Vail` treated only as a likely codename)
+- Silicon Motion **SM8366**
+- InnoGrit **Tacoma IG5669**
+
+The goal is **not** to reverse-engineer vendor firmware. The goal is to extract a credible public capability superset and turn it into a reusable controller model that better reflects modern enterprise SSD behavior.
+
+The target simulator should support:
+
+- multiple enterprise controller personas rather than one default controller shape
+- stronger queue arbitration, namespace isolation, and QoS behavior
+- controller modes such as conventional NVMe, ZNS, and FDP-oriented behavior
+- tighter coupling between controller profile and NAND/media profile
+- richer telemetry, security lifecycle, and maintenance interference modeling
+
+This design intentionally stays at the **controller / host-visible behavior** level. It does not attempt vendor-private opcode fidelity or cycle-accurate silicon modeling.
+
+---
+
+## 2. Current State
+
+### 2.1 What exists today
+
+HFSSS already has a useful controller/NVMe foundation:
+
+- controller context and configuration in [`include/controller/controller.h`](../../../include/controller/controller.h) and [`src/controller/controller.c`](../../../src/controller/controller.c)
+- arbitration and scheduling skeletons in [`include/controller/arbiter.h`](../../../include/controller/arbiter.h) and [`src/controller/scheduler.c`](../../../src/controller/scheduler.c)
+- DWRR-style QoS structures in [`include/controller/qos.h`](../../../include/controller/qos.h)
+- security primitives in [`include/controller/security.h`](../../../include/controller/security.h)
+- telemetry building blocks in [`include/common/telemetry.h`](../../../include/common/telemetry.h)
+- NVMe opcode and register surface in [`include/pcie/nvme.h`](../../../include/pcie/nvme.h)
+- user-space NVMe device emulation in [`src/pcie/nvme_uspace.c`](../../../src/pcie/nvme_uspace.c)
+
+### 2.2 Current limitations
+
+The current controller shape is still much closer to a single generic SSD model than to an enterprise controller family:
+
+- no explicit **controller profile** abstraction
+- no first-class distinction between 8-channel, 16-channel, or 16/18-channel controller personas
+- no **dual-port** host model
+- no PF/VF or tenant-partitioning model
+- scheduler behavior is materially simpler than public enterprise QoS / arbiter claims
+- no controller-level **ZNS / FDP / SEF / Open Channel** persona split
+- no controller-level **atomic write** or offload abstraction
+- telemetry, QoS, security, and NAND/media timing are present as separate subsystems rather than one coherent enterprise-controller behavior
+
+### 2.3 Related design work already in repo
+
+This design should be read alongside:
+
+- [`2026-04-10-nand-media-command-coverage-design.md`](2026-04-10-nand-media-command-coverage-design.md)
+
+The NAND/media design handles command-state and media-side realism. This controller design handles **host-facing policy, persona, and enterprise feature behavior**.
+
+---
+
+## 3. Public Capability Baseline
+
+### 3.1 Cross-vendor common signals
+
+Across the three public controller families, the following common signals are strong enough to drive a simulator design:
+
+- PCIe **Gen5 x4** is the baseline host link
+- high NAND parallelism is normal: **8 / 16 / 16-18 channels**
+- enterprise controllers expose stronger **QoS, latency isolation, and namespace behavior**
+- media and controller behavior are increasingly coupled through **high-speed ONFI/Toggle timing**
+- enterprise deployment modes now routinely include **ZNS** and other placement-aware behavior
+- security is controller-facing, not just media-facing:
+  - secure boot
+  - root of trust
+  - encryption lifecycle
+  - firmware protection
+- telemetry and observability are expected controller features
+
+### 3.2 Vendor-specific emphasis
+
+- **Marvell Bravera SC5**
+  - strongest inputs: 8/16-channel split, dual-port, PF/VF-like partitioning, SLA/QoS, SEF/ZNS/Open Channel flexibility
+
+- **Silicon Motion SM8366**
+  - strongest inputs: 16-channel enterprise profile, 1024 queue pairs, 128 namespaces, PerformaShape, FDP, host-based FTL cooperation
+
+- **InnoGrit Tacoma IG5669**
+  - strongest inputs: 16/18-channel coupling to NAND/media profile, ZNS, atomic write, low-latency persona, optional offload behavior
+
+### 3.3 What is explicitly out of scope
+
+Public documents do **not** reveal enough detail to justify:
+
+- vendor-private registers or admin opcodes
+- exact telemetry page formats
+- exact FTL algorithms
+- exact latency distributions under every workload
+- private firmware recovery state machines
+
+Those should remain abstract or deferred.
+
+---
+
+## 4. Goals And Non-Goals
+
+### 4.1 Goals
+
+- introduce a first-class **controller profile** abstraction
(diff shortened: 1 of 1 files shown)


AlvinYangZF:
## Deep review — feasibility, technical completeness, internal logic

Reviewed against master `4155575`, comparing every claim in the spec against actual code shape and against NVMe 2.x / OCP Datacenter NVMe SSD Spec coverage expectations.

**Three-axis verdict**:

| Axis | Result |
|---|---|
| Feasibility on top of master | Implementable **with major rewrites** |
| Technical completeness (NVMe 2.x / PCIe Gen5 / enterprise problem space) | **Major gaps** |
| Internal logical consistency | Mostly consistent, **with named issues** |

The architecture sketch (layer model, §5.6 precedence, §6.4 dual-port path states, phase plan backbone) is genuinely thoughtful. The problem is that a load-bearing share of the named capabilities have no implementation path on the current codebase, no normative reference to lean on, and a handful of phase-ordering contradictions that prevent phases from exiting.

---

### Must-fix before implementation can start

**CRITICAL feasibility blockers** (every one of these is prerequisite work the spec does not budget):

1. **Multi-namespace dispatch does not exist.** `src/pcie/nvme_uspace.c` hard-wires `if (nsid != 1) return error` on every read/write/trim/format/identify path. The FTL-side `ns_mapping_mgr` supports 32 namespaces but NVMe dispatch never reaches it. §3.2 persona claims 128 namespaces, §5.3 namespace-mode matrix, §6.5 ZNS+FDP coexistence, and §9.2 validation must-haves all assume multi-NS behavior that is currently impossible. **Prerequisite PR required** before Phase 2 can begin.

2. **Controller lifecycle state machine does not exist.** `grep "controller_state\|CTRL_BOOTSTRAP"` is zero hits. `struct controller_ctx` has only `bool running` / `bool initialized`. §5.6 makes the lifecycle layer the **top-precedence safety gate** for admission, command legality, telemetry visibility, and security-sensitive commands. Phases 2 and 3 silently depend on this gate, but the phase plan schedules it in Phase 4. Honest build order is **Phase 0 → Phase 1 → a lifecycle skeleton from Phase 4 → Phase 2 → Phase 3**, not the order currently written.

3. **ZNS and FDP are 100% greenfield.** Zero source-level scaffolding: no `NVME_CSS_ZNS`, no zone descriptors, no write pointer, no FDP Placement Identifier / Reclaim Group / Reclaim Unit Handle. Phase 2's "conventional / ZNS / FDP contract" bullet hides realistically **3000–6000 LOC of new code** plus test harness. This is by far the biggest under-estimation in the doc.

4. **1024 queue pairs vs 64 today (ABI break).** `struct nvme_controller_regs` has `doorbell[64]` literally in the packed layout; `BAR0_SIZE = 0x4000` cannot accommodate 1024 QP doorbells. Raising this is ABI-breaking for every existing NVMe test. §3.2 persona declares 1024 QP without flagging the structural rewrite.

5. **`hfsss_config` loader cannot express profile data.** Header comment says "no arrays, no multi-line strings, flat key/value." §6.1's `namespace_mode_matrix`, `telemetry_families[]`, `feature_bitmap`, and `command_set_bitmap` all need list/array/nested syntax that the current loader does not support. Either swap in libyaml, or extend the hand-rolled parser.

**HIGH-severity completeness gaps** (the spec needs to name what it covers, even if content is deferred):

6. **Zero normative standards references.** §12 cites only six vendor marketing URLs. A serious enterprise controller simulator spec has to anchor every "what does this field mean" decision to a versioned normative document. Add a References section listing:
   - NVM Express Base Specification (target revision, e.g., 2.0c or 2.1)
   - NVM Express NVM / Zoned / Key-Value Command Set Specifications
   - NVM Express Management Interface
   - **OCP Datacenter NVMe SSD Specification** (OCP conformance is almost always the acceptance gate for enterprise customers — the doc should declare whether it is a target)
   - TCG Storage Architecture Core / Opal SSC / Pyrite / Ruby / Enterprise SSC
   - PCI Express Base Specification 5.0
   - JEDEC JESD230 and/or ONFI 5.x for the media side
   - An explicit statement of **which revision is the baseline**, so implementation disputes resolve deterministically

7. **NVMe 2.x four-level hierarchy (Domain → Endurance Group → NVM Set → Namespace) is absent.** §5 layer model collapses to "namespace." FDP Reclaim Groups belong at the Endurance Group level. ANA belongs at the Domain level. "Multi-domain partitioning" is used four times without a fixed referent. Without this hierarchy, Phase 2's FDP contract and Phase 5's "multi-domain" work cannot be written.

8. **Command set and admin command scope not enumerated.** §6.1 lists `command_set_bitmap` as a field name. There is no table of which NVMe command sets are in scope, which admin commands (Identify CNS values, Get/Set Features, Get Log Page LIDs, Namespace Management, Directives, Virtualization Management, Sanitize, Format NVM, Firmware Commit/Download, Device Self-Test, NVMe-MI, Security Send/Receive), and which I/O commands (Flush, Read, Write, Compare, Write Zeroes, Dataset Management, Verify, Reservations, Copy, Zone Management Send/Receive, Zone Append). Phase 1 cannot populate `command_set_bitmap` without this table.

9. **ZNS and FDP treated symmetrically when they are architecturally asymmetric.** §5.3 lumps them together as "namespace modes." In NVMe, **ZNS is a separate Command Set** with its own Zone state machine (Empty → IOpened/EOpened → Closed → Full → Read Only → Offline), MAR/MOR limits, Zone Append, and error taxonomy. **FDP is a Directive on the NVM Command Set** with Placement Identifiers, Reclaim Groups, Reclaim Unit Handles (and RUH Type), FDP Events, and the LID 20h FDP Configurations log. Split §5.3 into two independent sub-contracts.

10. **Enterprise observability not mapped to NVMe/OCP log pages.** §9.4 schema fields (`profile_id`, `path_state`, `secure_boot_state`, `firmware_state`, `latency_histogram_window`, etc.) are hand-rolled. NVMe already defines SMART/Health, Firmware Slot Info, Error Information, Sanitize Status, Persistent Event Log, Telemetry Host/Controller-Initiated, Latency Monitor (Feature 0Dh + log page), Endurance Group Information. OCP Datacenter NVMe SSD Spec defines C0h–CAh with Error Recovery, FW Activation History, Latency Monitor, Device Capabilities, Unsupported Requirements, TCG Configuration, Telemetry String, SMART per-namespace. **Not one of these is referenced.** Add an explicit mapping table from §9.4 observables to their target LID, and declare whether OCP conformance is in scope.

---

### Internal logic issues that block phase exits

11. **Circular dependency: Phase 3 exit requires telemetry reflects path state, but path-state model is Phase 4.**
    - §8 Phase 3 exit: *"telemetry reflects queue pressure, latency, maintenance state, **and path state**"*
    - §8 Phase 4 deliverables: *"dual-port host model / **path-state model**"*
    - As written, **Phase 3 cannot exit**. Plus §5.2 lists "port/path management" as a Phase 3 arbitration domain, compounding the problem. Fix: either move a minimal path-state stub to Phase 3, or remove "path state" from the Phase 3 exit criterion.

12. **Phase 1 declares capability flags that have no semantics until Phase 5.** §6.1 Phase 1 profile fields include `supports_atomic_write`, `supports_multi_domain_partitioning`, `supports_offload`. §8 Phase 5 deliverables list the *capability models* behind those flags. §8 Phase 1 exit says *"enterprise capability flags come from the active profile"* as if they are meaningful at Phase 1. Either (a) state explicitly that Phase 1 ships flag *slots* with no enforced semantics, (b) move the flag definitions to their implementing phase, or (c) reclassify them as profile metadata. Same issue affects `supports_dual_port` (Phase 1 vs 4) and `security_lifecycle_class` (Phase 1 vs 4).

13. **"Controller modes" vs "namespace-scoped modes" contradiction.** §1 summary calls ZNS and FDP *"controller modes"*. §6.2 and §6.5 and §7.3 all insist they are *"namespace-scoped, not controller-wide"* and explicitly forbid *"device-wide `zns` or `fdp` controller identities."* §1's wording is rejected by every subsequent normative section.

14. **`atomic write` has no layer owner.** §5.3: *"atomic write is **not** owned by this layer."* §6.6 lists *"atomic-write completion/failure semantics"* under "controller-owned contracts" without naming a layer. No other section claims it. Phase 5 lists an "atomic write capability model" with no home. Name the owner, or explicitly declare it a cross-cutting concern bound to a specific layer (profile or lifecycle are the reasonable choices).

15. **Terminology drift: persona / profile / preset.** §6.1 calls the struct a "profile" with `profile_name` and `profile_version` fields. §6.2 calls the instances "personas." §8 calls them "persona presets." §5.1 describes a three-axis decomposition; §8 line 541 describes a five-axis decomposition. Pick one noun for the struct, one noun for a named instance, and one axis list. Three-way overlap is not reconcilable without a definition.

---

### Temporal dependency on NAND/media work

The §5.4 Media Coupling Layer names five contracts: timing vectors, legality/capability feed, suspend/status snapshot, reset-abort propagation, media-error propagation. Those primitives live in the NAND/Media command engine work (spec #71, skeleton implementation in flight at #73). On master, `include/media/cmd_engine.h` does not exist; the #73 Phase 1 skeleton implements only the die state machine and snapshot surface, **not** the negotiation or legality feed that §5.4 expects to consume. Any Phase 2+ controller work started before the NAND/media engine reaches its own Phase 3 will either block or build against a floating branch.

Recommendation: **cite the NAND/media phase plan explicitly in §5.4** and pin a hard requirement that controller phase N depends on NAND/media phase M, so the work cannot be scheduled out of order.

---

### Suggested spec edit plan

Before this design is accepted as an implementation contract, I believe the following edits are needed:

- Insert **Phase 0.5: prerequisite plumbing** covering multi-NS dispatch (Blocker 1), a minimal controller lifecycle skeleton (Blocker 2), doorbell/QP capacity expansion (Blocker 4), and an array-capable config loader (Blocker 5)
- Split the **controller lifecycle skeleton** out of Phase 4 and move it earlier, because §5.6 makes it a Phase 2/3 dependency
- Move a minimal **path-state stub** into Phase 3 to break the circular exit criterion
- Replace §12 with **normative references** (NVMe Base 2.0c or 2.1, NVMe NVM/ZNS/KV CS, NVMe-MI, OCP Datacenter NVMe SSD Spec, TCG Storage Core / Opal / Pyrite / Ruby / Enterprise SSC, PCIe Base 5.0, JEDEC/ONFI) and declare the baseline revision
- Add a **Domain → Endurance Group → NVM Set → Namespace hierarchy section** mapping each level to its NVMe identifier and to the enterprise capability it owns
- Add a **command set and admin command scope table** marking each command baseline / optional / deferred
- **Split §5.3 into two sub-contracts**: §5.3.1 ZNS (separate Command Set with zone state machine) and §5.3.2 FDP (Directive on NVM CS with Reclaim Groups)
- Add an **observability mapping table** from §9.4 schema fields to NVMe log page LIDs and OCP C0h–CAh equivalents, and declare whether OCP Datacenter NVMe SSD Spec conformance is in scope
- Expand §4.2 non-goals to explicitly name items currently silent: PCIe ASPM/L1.x/DSTATE, NVMe Power States and APST, HMB/PMR/CMB, T10 DIF/DIX end-to-end data protection, Reservations, ANA groups, Namespace Management admin command, SR-IOV primary/secondary controller admin surface, Device Self-Test, Persistent Event Log, firmware slots and FW Commit actions, Sanitize variants, TCG Opal locking ranges, NVMe-MI side-band. Each needs an explicit in-scope / out-of-scope call-out rather than silent absence.
- **Lock terminology**: one word for the struct, one for an instance. Reconcile the three-axis and five-axis profile decompositions.
- **Name an owner** for atomic write, `supports_multi_domain_partitioning`, `supports_offload`, and any other Phase-1-declared-but-Phase-5-delivered flag.

Happy to review the revised version.

PR #73, opened 2026-04-11
feat: NAND command engine skeleton (design spec phase 1)

AlvinYangZF:
## Summary

Implements the first phase of the NAND/media command coverage design merged in #71. Introduces an explicit per-die command state machine and an in-flight command-state object behind the existing \`media_nand_read/program/erase\` entry points, without changing any public API or observable aggregate latency.

Three new headers and four new sources form a small, decoupled engine under \`src/media/\`:

- **cmd_state** — 11 die states, 5 command phases, \`nand_die_cmd_state\` embedded per die, and a dedicated per-die lock
- **cmd_legality** — static \`[state][opcode]\` matrix plus accept/complete/abort transition helpers
- **cmd_engine** — synchronous submit/snapshot/reset API driven by a \`nand_cmd_ops\` callback vtable; commit hooks own storage mutation, engine owns legality + state transitions + stage-aware EAT updates
- **cmd_stage_timing** — stage-budget partition; the full latency currently pins onto the array-busy stage to preserve wrapper-compatible aggregate timing

Existing paths are rewired through the engine:

- \`media_nand_read/program/erase\` keep byte-identical public signatures and return codes
- read state validation runs in \`on_setup_commit\` so a failed read against an unwritten page consumes zero EAT (preserves the legacy short-circuit)
- array-stage EAT is committed before the commit hook runs so the caller-visible completion boundary matches the legacy eat-before-mutate ordering
- \`nand_die\` gains \`cmd_state\` + \`die_lock\`; initialization wired through \`nand_init_hierarchy\` / \`nand_cleanup_hierarchy\`
- persistence format in \`media_save/load\` audited — untouched (new fields are in-memory only)

### Design alignment

- §5 three-layer split, §7.1 11 die states, §7.2 command context field set (all 9 spec fields present), §7.4 die-level lock protects the command-state object, §11 deliverables for the skeleton, §12.2 plane mask u32

## Exit criteria status

All exit criteria from §11 phase 1 satisfied:

- [x] \`media_nand_*\` entry points flow through the new engine
- [x] Functional compatibility: data correctness, page/block state transitions, error handling equivalent
- [x] Timing compatibility: aggregate latency wrapper-compatible (sub-microsecond tolerance)
- [x] \`include/media/media.h\` byte-identical to master (confirmed via \`git diff\`)
- [x] \`tests/test_media.c\` 116/116 (added 40 state-machine assertions)
- [x] \`tests/test_hal.c\` 61/61
[...]

PR #74, opened 2026-04-12
feat: NAND command engine phase 2 foundational standard commands

AlvinYangZF:
## Summary

Implements **Phase 2** of the NAND/Media command coverage design spec (`docs/superpowers/specs/2026-04-10-nand-media-command-coverage-design.md` §11). Builds on the Phase 1 engine skeleton merged in #73.

New foundational standard commands exposed through both the internal engine and the public media API:

- **read_status** — 1-byte classic ONFI status (RDY / ARDY / WP_N / FAIL)
- **read_status_enhanced** — structured decode over the in-flight command-state snapshot
- **read_id** — 8-byte ID with ONFI 4.2 Table 37 bit layout in bytes [0..4] and Toggle-extended reserved bytes [5..7]
- **read_parameter_page** — minimum §6.3 contract: identity, geometry, addressing, ECC advertisement, supported-command bitmap, timing advertisement, revision/size prefix, CRC-16-CCITT over stable prefix

Identity and parameter page live on `struct nand_device` and are populated in `media_init` after the timing/EAT wiring so `tR` / `tPROG` / `tBERS` advertised values reflect the active timing model.

## Critical review fixes included

During internal code review these issues were found and fixed in the same branch before the initial PR:

1. **CRITICAL — identity-during-suspend destroyed suspend context.** The initial implementation routed identity ops through `nand_cmd_state_begin` which clobbered `target`, `remaining_ns`, `suspend_count`, `in_flight`, and forced `DIE_IDLE`. Legality allows identity ops during `DIE_SUSPENDED_*` per ONFI, so this would have silently broken Phase 3 resume. Fixed by simplifying `engine_submit_identity` to a pure payload copy that never touches `cmd_state`.

2. **CRITICAL — status observability hole.** `engine_submit` in Phase 1 held `die_lock` across the entire submission (busy-wait + commit hooks), so concurrent `read_status` callers would block until the command completed and could only observe settled states. This violated the design spec concurrency contract. Fixed by narrowing `die_lock` to short transactional segments around `cmd_state` updates and releasing it across the busy-wait and commit callbacks. Channel-level serialization still comes from `channel_lock`.

3. **HIGH — FAIL latch semantics.** ONFI 4.2 Status Register requires FAIL to latch until the next PROG/ERASE completes successfully or a RESET occurs. The initial implementation derived FAIL from the last op's `result_status` which was cleared by any intervening read or identity op. Added a sticky `latched_fail` field on `nand_die_cmd_state`; set on PROG/ERASE failure completion, cleared on successful PROG/ERASE completion and on RESET, untouched by READ / READ_ID / READ_STATUS / READ_PARAM_PAGE.

4. **HIGH — addr_cycles_row computation.** Dead `if/else` branches, self-overwriting assignment, and missing `luns_per_ce` multiplier under-reported row-address cycles on multi-LUN parts. Replaced with a single `u64` computation across `pages_per_block × blocks_per_lun × luns_per_ce`.

## Test coverage

New test cases `SM-14` .. `SM-24` in `tests/test_media.c`:

| Case | Focus |
|------|-------|
| SM-14 | Status byte on fresh idle die |
| SM-15 | Status byte across a completed program |
| SM-16 | Enhanced status decode post-reset |
| SM-17 | Exhaustive legality matrix for status and identity opcodes |
| SM-18 | read_id structural correctness with TLC cell-type field |
| SM-19 | Parameter page minimum field contract |
| SM-20 | Parameter page stability across consecutive reads |
| SM-21 | Reset produces deterministic clean baseline with no EAT drift |
| SM-22 | Invalid-argument rejection paths |
| SM-23 | FAIL latch persistence across read and read_id, cleared by reset and successful program |
| SM-24 | Identity ops during synthetic `DIE_SUSPENDED_PROG` preserve state, phase, opcode, target, in_flight, remaining_ns, suspend_count |

## Test plan

- [x] \`test_media\` 243/243 (126 Phase 1 + 88 Phase 2 + 29 review-fix)
- [x] \`test_hal\` 61/61
- [x] \`test_reliability\` 76/76
- [x] \`test_ftl_reliability\` 31/31
- [x] \`test_mt_ftl\` 10/10 (0 TAA conflicts)
- [x] \`systest_data_integrity\` 50/50 (400.8 s)
- [x] \`systest_wear_gc\` 48/48 (1456.7 s)
- [x] Clean rebuild after clang-format, no new warnings

### DIFF
--- a/include/media/cmd_engine.h
+++ b/include/media/cmd_engine.h
@@ -3,6 +3,7 @@
 
 #include "common/common.h"
 #include "media/cmd_state.h"
+#include "media/nand_identity.h"
 
 struct nand_device;
 struct eat_ctx;
@@ -35,6 +36,28 @@ int nand_cmd_engine_submit_reset(struct nand_device *dev, const struct nand_cmd_
 
 int nand_cmd_engine_snapshot(struct nand_device *dev, const struct nand_cmd_target *target,
                              struct nand_die_cmd_state *out);
+
+/*
+ * Status submission paths. Both take only the die-level lock so they remain
+ * observable mid-flight; neither touches the channel lock or the cmd_state
+ * begin/advance path.
+ */
+int nand_cmd_engine_submit_read_status(struct nand_device *dev, const struct nand_cmd_target *target,
+                                       u8 *out_status_byte);
+
+int nand_cmd_engine_submit_read_status_enhanced(struct nand_device *dev, const struct nand_cmd_target *target,
+                                                struct nand_status_enhanced *out);
+
+/*
+ * Identity submission paths. Both take the full channel+die lock order and
+ * are legal only from DIE_IDLE or the suspend states (non-conflicting reads).
+ * Current phase skips EAT advancement for these bus-bound operations — see
+ * engine_submit_identity in cmd_engine.c.
+ */
+int nand_cmd_engine_submit_read_id(struct nand_device *dev, const struct nand_cmd_target *target, struct nand_id *out);
+
+int nand_cmd_engine_submit_read_param_page(struct nand_device *dev, const struct nand_cmd_target *target,
+                                           struct nand_parameter_page *out);
 
 /*
  * Stage-budget split. Kept in its own translation unit so the partition
--- a/include/media/cmd_state.h
+++ b/include/media/cmd_state.h
@@ -32,7 +32,17 @@ enum nand_cmd_phase {
  * Engine-level opcode enum. Distinct from the hardware-facing enum nand_cmd
  * in nand.h which encodes raw NAND command bytes.
  */
-enum nand_cmd_opcode { NAND_OP_READ = 0, NAND_OP_PROG, NAND_OP_ERASE, NAND_OP_RESET, NAND_OP_COUNT };
+enum nand_cmd_opcode {
+    NAND_OP_READ = 0,
+    NAND_OP_PROG,
+    NAND_OP_ERASE,
+    NAND_OP_RESET,
+    NAND_OP_READ_STATUS,
+    NAND_OP_READ_STATUS_ENHANCED,
+    NAND_OP_READ_ID,
+    NAND_OP_READ_PARAM_PAGE,
+    NAND_OP_COUNT
+};
 
 struct nand_cmd_target {
     u32 ch;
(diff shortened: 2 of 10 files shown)


AlvinYangZF:
High: the new parameter-page payload is ABI-dependent, so the same simulated device can expose different bytes and CRCs across builds. `struct nand_parameter_page` is returned as a raw native C layout in `include/media/nand_identity.h`, and `size_bytes` / `crc` are derived from `sizeof(struct nand_parameter_page)` and `offsetof(..., crc)` in `src/media/nand_identity.c`. Because the struct is not packed and contains naturally aligned integer fields, the serialized blob becomes compiler/ABI sensitive. That conflicts with the PR’s stated goal of a stable `read parameter page` contract.

AlvinYangZF:
High: `RESET` still does not satisfy the design contract this PR claims to implement. The engine holds `channel->lock` for the full synchronous submit path in `src/media/cmd_engine.c`, and `nand_cmd_engine_submit_reset()` explicitly notes that true mid-flight abort is deferred because reset only runs after the active command drains. But the Phase 2 NAND/media design says reset should abort in-flight commands and return the die to a clean state. The implementation is internally consistent, but it does not currently match the spec contract advertised by the PR.

AlvinYangZF:
Medium-high: the parameter page under-reports required row/column address cycles for non-power-of-two geometries. `src/media/nand_identity.c` computes `row_bits` / `col_bits` with floor-log2 loops and then converts those to byte counts. For values just above a byte boundary, that is wrong: for example, 257 rows or columns still yields 1 cycle here, but it needs 2. That makes the advertised parameter page self-inconsistent on larger or irregular geometries.

AlvinYangZF:
Medium: `remaining_ns` in `read_status_enhanced` is stale for the entire lifetime of an in-flight command. It is initialized once in `src/media/cmd_state.c`, then exposed directly in `src/media/nand_identity.c`, but nothing in the command engine ever decrements it; the only later write is resetting it to zero at completion in `src/media/cmd_engine.c`. Any caller using enhanced status for progress, suspend timing, or future resume decisions will see the original full budget instead of actual remaining time.

AlvinYangZF:
Medium: the new tests overstate what they prove for the two highest-risk fixes. The FAIL-latch and suspend-preservation cases in `tests/test_media.c` preload internal state by mutating `die0->cmd_state` directly, then verify that status/identity reads preserve those fields. That does not prove the engine can actually enter, preserve, clear, or observe those states through real command submission paths. It also leaves the claimed mid-flight observability / locking behavior unpinned by any true concurrent test.

PR #75, opened 2026-04-12
feat: NAND command engine suspend/resume

AlvinYangZF:
## Summary

- Implement program suspend/resume and erase suspend/resume per design spec §8
- Release channel lock during array-busy spin, enabling concurrent suspend/reset via atomic interrupt flags
- Read-during-suspend conflict gate: same-page/block returns BUSY, non-conflicting reads preserve all suspension context
- Reset force-cleans state under die_lock without worker-wait, worker skips final block on abort
- Add tSSBSY/tRSBSY timing constants per cell family, charged to both EAT and wall-clock
- Fix ARDY status byte to report 0 during suspended states (ONFI 4.2 compliance)
- 67 new test assertions (SM-25..SM-34) using pthread worker pattern for real concurrent suspend/resume

## Review fixes applied pre-merge

| ID | Severity | Fix |
|----|----------|-----|
| R4 | CRITICAL | Recheck abort_request inside die_lock window before writing suspended state |
| R5 | MEDIUM | Charge tSSBSY to EAT via eat_update_stage in suspend arm |
| R6 | MEDIUM | Added SM-32 (erase reset-abort symmetric test) |
| R3 | HIGH | Documented u64 alignment assumption for x86-64/AArch64 targets |

## Regression results

| Suite | Result |
|-------|--------|
| test_media | 310/310 |
| test_hal | 61/61 |
| test_reliability | 76/76 |
| test_mt_ftl | 10/10 |

## Test plan

- [ ] CI clang-format check
- [ ] CI ASan / TSan / UBSan
- [ ] CI system tests (systest_data_integrity, systest_wear_gc)
- [ ] Verify SM-25..SM-34 pass with TSan enabled
- [ ] Review conflict gate for future multi-plane extension (deferred to Phase 4)

### DIFF
--- a/include/media/cmd_engine.h
+++ b/include/media/cmd_engine.h
@@ -60,6 +60,19 @@ int nand_cmd_engine_submit_read_param_page(struct nand_device *dev, const struct
                                            struct nand_parameter_page *out);
 
 /*
+ * Suspend/resume submission paths. All four take only the die-level lock
+ * so the submitting thread can reach the running worker without blocking
+ * on the channel lock that the worker released across its array-busy
+ * spin. The suspend entry points only use (ch, chip, die) from target;
+ * block/page/plane_mask fields are ignored because these commands act on
+ * the currently in-flight op of the addressed die.
+ */
+int nand_cmd_engine_submit_prog_suspend(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_prog_resume(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_erase_suspend(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_erase_resume(struct nand_device *dev, const struct nand_cmd_target *target);
+
+/*
  * Stage-budget split. Kept in its own translation unit so the partition
  * can be refined later without touching the engine.
  */
--- a/include/media/cmd_state.h
+++ b/include/media/cmd_state.h
@@ -1,6 +1,8 @@
 #ifndef __HFSSS_NAND_CMD_STATE_H
 #define __HFSSS_NAND_CMD_STATE_H
 
+#include <stdatomic.h>
+
 #include "common/common.h"
 #include "common/mutex.h"
 
@@ -41,6 +43,10 @@ enum nand_cmd_opcode {
     NAND_OP_READ_STATUS_ENHANCED,
     NAND_OP_READ_ID,
     NAND_OP_READ_PARAM_PAGE,
+    NAND_OP_PROG_SUSPEND,
+    NAND_OP_PROG_RESUME,
+    NAND_OP_ERASE_SUSPEND,
+    NAND_OP_ERASE_RESUME,
     NAND_OP_COUNT
 };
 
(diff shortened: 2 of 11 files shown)


AlvinYangZF:
High: the current `read-during-suspend` path is still serialized by the original `PROG/ERASE` EAT window, so it does not actually execute in the suspended window.

In `engine_submit()`, the worker books the full `array_ns` into EAT before entering the suspendable array-busy loop (`src/media/cmd_engine.c`). Later, the non-conflicting read fast path still calls `engine_wait_until_available()` before doing the read work (`src/media/cmd_engine.c`). Because `eat_get_max()` is shared across READ / PROGRAM / ERASE resources (`src/media/eat.c`), that read still waits behind the original program/erase deadline even after the die has transitioned to `DIE_SUSPENDED_*`.

That means the implementation currently provides *legality* for read-during-suspend, but not the intended *concurrency*. The design spec says a legal read during suspend should be able to run in the suspended window. As written, the code preserves the old serialization point instead.

I would treat this as a real design/implementation mismatch, not just a timing-detail nit.

AlvinYangZF:
High: `RESET` is still not a full abort of an in-flight `PROG/ERASE`.

The reset path force-cleans `cmd_state` and sets `abort_request` under `die_lock` (`src/media/cmd_engine.c`), but the worker only samples `abort_request` inside `engine_run_array_busy()`. The setup hook runs earlier, and the array commit hook runs later, outside that abort sampling window.

So there are still race windows where:
- reset lands after submit has begun but before the array-busy loop starts, or
- reset lands after the array-busy loop returns but before `on_array_commit` executes

In those windows, the worker can still execute commit logic after reset has already reported success and published a clean state. That does not satisfy the design contract that reset aborts any in-flight command and prevents the command from committing.

The current tests mostly cover `reset-after-suspend-ack`, which is a narrower case than the full abort contract the PR claims to implement.

AlvinYangZF:
Medium-high: reset does not restore a fully clean observable command-state snapshot.

`nand_cmd_state_init()` clears opcode, target, timestamps, suspend counters, and suspend bookkeeping (`src/media/cmd_state.c`), but `nand_cmd_engine_submit_reset()` only clears a subset of those fields (`src/media/cmd_engine.c`). As a result, post-reset status can still retain stale values for fields such as:
- `opcode`
- `target`
- `start_ts_ns`
- `phase_start_ts_ns`
- `suspend_count`
- `last_suspend_ts_ns`

Those fields are then surfaced by `nand_status_enhanced_from_cmd_state()` (`src/media/nand_identity.c`). So after a reset, the die may be reported as `IDLE`, but still leak prior suspend/program history through the enhanced status payload.

If the intended contract is "reset returns the die to a clean state", I think this needs to be normalized to the same baseline as `nand_cmd_state_init()` or another explicitly defined post-reset canonical state.

AlvinYangZF:
Medium: resume overhead is charged to EAT, but not to wall-clock latency.

The PR summary says `tSSBSY/tRSBSY` are charged to both EAT and wall-clock. The suspend side does that: after entering the suspended state, the code both updates EAT and busy-spins for `tSSBSY` (`src/media/cmd_engine.c`).

But the resume path is asymmetric. `engine_submit_resume()` only calls `eat_update_stage()` for `tRSBSY` and returns (`src/media/cmd_engine.c`). There is no matching wall-clock drain. Then the worker resets `array_started_ns` immediately after noticing the resumed state.

So scheduler accounting sees resume overhead, but caller-visible latency does not. That is inconsistent with both the PR summary and the design requirement that suspend/resume overhead affect observable latency, not just EAT statistics.

AlvinYangZF:
Medium: the new tests miss the two highest-risk behavioral gaps in this PR.

First, the reset-abort tests (`SM-30`, `SM-32`) only issue reset after suspend has already been acknowledged in `tests/test_media.c`. That proves `reset-while-suspended`, but it does not prove the stronger race the implementation claims to support: reset preempting the live array-busy worker and preventing later commit.

Second, the read-during-suspend coverage (`SM-28`) only proves that a non-conflicting read eventually succeeds and preserves suspend context. It does *not* assert that the read completed inside the suspended window rather than being serialized behind the original program/erase EAT deadline.

So the suite is good at checking state preservation and legality, but it still does not pin down:
- true mid-flight reset-abort semantics
- true suspended-window read concurrency

Given the complexity of the locking/timing changes in this PR, I think those missing assertions matter.

PR #76, opened 2026-04-12
feat: multi-plane read/program/erase simulation

AlvinYangZF:
## Summary

- Lift single-plane guard in command engine: `plane_mask` now supports multiple set bits
- Multi-plane EAT charging: wait for max deadline across all planes, charge each stage to all planes simultaneously (ONFI 4.2 die-internal parallelism)
- Three new media-layer multi-plane wrappers with per-plane data/spare arrays and atomic commit hooks
- Plane legality validation: geometry bounds, `enable_multi_plane` config gate, bad-block atomicity across planes
- Read-during-suspend conflict check upgraded to use plane_mask intersection
- Addresses PR #73 reviewer feedback on single-plane target model ABI

## Depends on

- PR #75 (suspend/resume) — this branch is based on Phase 3

## Regression results

| Suite | Result |
|-------|--------|
| test_media | 344/344 |
| test_hal | 61/61 |
| test_reliability | 76/76 |
| test_mt_ftl | 10/10 |

## Test plan

- [ ] CI clang-format check
- [ ] CI ASan / TSan / UBSan
- [ ] CI system tests
- [ ] Verify SM-35..SM-50 pass with TSan enabled
- [ ] Validate multi-plane EAT matches single-plane timing (SM-43)
- [ ] Validate bad-block atomicity (SM-42)

### DIFF
--- a/include/media/cmd_engine.h
+++ b/include/media/cmd_engine.h
@@ -60,6 +60,19 @@ int nand_cmd_engine_submit_read_param_page(struct nand_device *dev, const struct
                                            struct nand_parameter_page *out);
 
 /*
+ * Suspend/resume submission paths. All four take only the die-level lock
+ * so the submitting thread can reach the running worker without blocking
+ * on the channel lock that the worker released across its array-busy
+ * spin. The suspend entry points only use (ch, chip, die) from target;
+ * block/page/plane_mask fields are ignored because these commands act on
+ * the currently in-flight op of the addressed die.
+ */
+int nand_cmd_engine_submit_prog_suspend(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_prog_resume(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_erase_suspend(struct nand_device *dev, const struct nand_cmd_target *target);
+int nand_cmd_engine_submit_erase_resume(struct nand_device *dev, const struct nand_cmd_target *target);
+
+/*
  * Stage-budget split. Kept in its own translation unit so the partition
  * can be refined later without touching the engine.
  */
--- a/include/media/cmd_state.h
+++ b/include/media/cmd_state.h
@@ -1,6 +1,8 @@
 #ifndef __HFSSS_NAND_CMD_STATE_H
 #define __HFSSS_NAND_CMD_STATE_H
 
+#include <stdatomic.h>
+
 #include "common/common.h"
 #include "common/mutex.h"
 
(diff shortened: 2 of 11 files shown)


AlvinYangZF:
Remaining high-severity issue after the latest fixes: `RESET` still races with a newly submitted command, so the old in-flight worker can miss the abort signal.

`nand_cmd_engine_submit_reset()` re-arms `abort_request`, but a new command immediately clears it again in `nand_cmd_state_begin()` (`src/media/cmd_state.c`). The old worker only samples that flag later in `engine_run_array_busy()` / the post-busy checks (`src/media/cmd_engine.c`). In the current ordering, reset is not yet a hard serialization point for the die: a reset can be accepted, a new command can begin, and the previous worker can then continue because the abort flag was cleared before it observed it.

I think this still falls short of the intended reset contract. The clean fix probably needs a generation/epoch-style abort token or some other mechanism that cannot be cleared by a subsequent command before the old worker has observed the reset.

AlvinYangZF:
`media_nand_multi_plane_read()` still regresses bad-block semantics compared with the single-plane API.

The single-plane path rejects bad blocks before submit (`media_nand_read()` in `src/media/media.c`), but `media_nand_multi_plane_read()` only validates geometry / page state and never performs the equivalent bad-block check. As a result, the multi-plane API can read through a block that is already marked bad, even though the single-plane API would reject it.

I would suggest aligning the multi-plane read wrapper with the existing single-plane contract before the command is submitted.

AlvinYangZF:
There is still an unchecked null-pointer path in `media_nand_multi_plane_read()`.

The wrapper validates that `data_array` itself is non-null, but it does not validate each `data_array[idx]`. Later, `mp_read_on_data_xfer_commit()` passes each per-plane pointer into the read/bit-error path, which eventually dereferences it. So `data_array != NULL` with one null entry is still a crashable input shape.

`media_nand_multi_plane_program()` already validates each entry; the read side should probably mirror that contract.

AlvinYangZF:
The new suspend/resume regression coverage still does not truly exercise a multi-plane in-flight command.

`SM-46` explicitly uses a single-plane worker on a 2-plane die, so it never covers suspend/resume behavior with `plane_mask != 1`. `SM-47` also contains a `media_nand_program()` call whose return value is overwritten immediately by the next call, so that branch is not actually asserted.

Given that Phase 4 is specifically about multi-plane command support, I think the highest-value follow-up test is a real in-flight multi-plane program/erase case that suspends, resumes, and then verifies both the legality path and the observable timing/commit behavior.

AlvinYangZF:
`SM-43` is still too weak to pin down the important EAT contract for multi-plane execution.

Right now the test only proves that both planes’ EAT values increased. It does not verify the critical behavior in the engine: waiting on the slowest prior plane / max prior deadline and then charging the selected planes with the same stage budget.

So a regression in the actual die-internal parallel timing semantics could still slip through while `SM-43` stays green. I think this test needs one more assertion around the relative deadlines, not just "both planes advanced".

PR #77, opened 2026-04-12
feat: cache read/program with stage-aware overlap

AlvinYangZF:
## Summary

- Implement cache read, cache read end, cache program (ONFI cache command pipeline)
- Die stays in mid-pipeline state (DATA_XFER / ARRAY_BUSY) between cache pages instead of returning to IDLE
- Cache continuation pages skip EAT wait for stage overlap
- Implicit cache sequence termination on non-cache command submission
- New timing params: tCBSY and tDCBSYR1 per cell family
- cmd_state tracks cache_active and cache_seq_count across sequence
- Three new media-layer wrappers

## Depends on

- PR #76 (multi-plane) → PR #75 (suspend/resume)

## Regression results

| Suite | Result |
|-------|--------|
| test_media | 408/408 |
| test_hal | 61/61 |
| test_reliability | 76/76 |
| test_mt_ftl | 10/10 |

## Test plan

- [ ] CI clang-format / cppcheck / TSan
- [ ] SM-56..SM-62 validate cache command data correctness and legality
- [ ] SM-57 validates cache program timing advantage over plain program
- [ ] Verify no regression in existing Phase 1-4 tests

### DIFF
--- a/include/media/cmd_engine.h
+++ b/include/media/cmd_engine.h
@@ -72,6 +72,13 @@ int nand_cmd_engine_submit_prog_resume(struct nand_device *dev, const struct nan
 int nand_cmd_engine_submit_erase_suspend(struct nand_device *dev, const struct nand_cmd_target *target);
 int nand_cmd_engine_submit_erase_resume(struct nand_device *dev, const struct nand_cmd_target *target);
 
+int nand_cmd_engine_submit_cache_read(struct nand_device *dev, const struct nand_cmd_target *target,
+                                      const struct nand_cmd_ops *ops, void *cb_ctx);
+int nand_cmd_engine_submit_cache_read_end(struct nand_device *dev, const struct nand_cmd_target *target,
+                                          const struct nand_cmd_ops *ops, void *cb_ctx);
+int nand_cmd_engine_submit_cache_program(struct nand_device *dev, const struct nand_cmd_target *target,
+                                         const struct nand_cmd_ops *ops, void *cb_ctx);
+
 /*
  * Stage-budget split. Kept in its own translation unit so the partition
  * can be refined later without touching the engine.
--- a/include/media/cmd_engine.h
+++ b/include/media/cmd_engine.h
@@ -83,6 +83,7 @@ int nand_cmd_engine_submit_cache_program(struct nand_device *dev, const struct n
  * Stage-budget split. Kept in its own translation unit so the partition
  * can be refined later without touching the engine.
  */
-void nand_cmd_stage_budget(enum nand_cmd_opcode op, u64 total_ns, u64 *setup_ns, u64 *array_ns, u64 *xfer_ns);
+void nand_cmd_stage_budget(enum nand_cmd_opcode op, u64 total_ns, u64 cache_overlap_ns, u64 *setup_ns, u64 *array_ns,
+                           u64 *xfer_ns);
 
 #endif /* __HFSSS_NAND_CMD_ENGINE_H */
--- a/include/media/cmd_state.h
+++ b/include/media/cmd_state.h
@@ -47,6 +47,9 @@ enum nand_cmd_opcode {
     NAND_OP_PROG_RESUME,
     NAND_OP_ERASE_SUSPEND,
     NAND_OP_ERASE_RESUME,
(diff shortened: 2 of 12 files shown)


AlvinYangZF:
`media_nand_cache_read()` and `media_nand_cache_program()` currently bypass the bad-block guardrails that the plain read/program wrappers already enforce.

In the plain paths, `media_nand_read()` rejects a bad block before submit, and `media_nand_program()` also fails closed on the `bbt_is_bad() == -1` error case. The new cache wrappers do not do the equivalent checks before they hand control to the command engine. That means the cache APIs can succeed on a block that the existing non-cache APIs would reject, which makes the Phase 5 surface semantically inconsistent with the pre-existing media contract.

I think the cache wrappers should preserve the same front-door validation behavior as their plain counterparts before they enter the engine.

AlvinYangZF:
The new cache timing knobs (`tCBSY` / `tDCBSYR1`) are currently dead code.

This PR adds the timing fields and getters, but the execution path still routes all cache commands through the same `nand_cmd_stage_budget()` partition where the full command budget goes to `array_ns`, and the engine never consumes `timing_get_cache_busy_ns()` or `timing_get_data_cache_busy_read_ns()`. So the overlap amount is not actually being driven by the new timing parameters.

That matters because the PR summary and Phase 5 design both claim stage-aware cache overlap. Right now the implementation gets its observable speedup from skipping the EAT wait on cache continuation, not from applying `tCBSY` / `tDCBSYR1` to the staged timing model. As written, changing those new timing values has no effect on behavior.

AlvinYangZF:
The mid-pipeline cache state is currently sticky rather than time-bounded.

After a successful `CACHE_READ`, the engine leaves the die in `DIE_READ_DATA_XFER`; after `CACHE_PROG`, it leaves the die in `DIE_PROG_ARRAY_BUSY`. But there is no worker or timer that naturally advances that state after return, and the new cache timing parameters are not being used to age it out. In practice, the die stays in that mid-pipeline state until the next command or reset forces a transition.

That means status/legality become command-driven rather than time-driven: if the host waits longer than any nominal cache-overlap window but submits nothing, the die still looks busy. For a cache-pipeline model, I think that is the wrong shape. The overlap window should be finite and derived from timing, not an indefinite state latch that only the next command clears.

AlvinYangZF:
`SM-56` does not currently protect the most important cache-read contract.

The test measures `plain_2_ns` and `cache_2_ns`, but then explicitly discards both values instead of asserting anything about them. So the suite verifies correctness of the returned data, but it does not pin down that cache read is observably distinct from plain read in the timing/resource model.

Given that Phase 5’s exit criterion is specifically that cache behavior be observably distinct from plain commands, I think this is a real coverage gap. As written, a regression that preserves the data path but loses the cache-read overlap semantics would still leave `SM-56` green.

AlvinYangZF:
Status: PR #77 review fixes complete, all tests passing locally.

  4 review comments addressed:

  1. Bad-block guardrails (media.c) — Added bbt_is_bad() checks to media_nand_cache_read, media_nand_cache_read_end, and media_nand_cache_program, matching the validation pattern of their plain counterparts.
  2. tCBSY/tDCBSYR1 no longer dead code (cmd_stage_timing.c, cmd_engine.c, cmd_engine.h) — nand_cmd_stage_budget() now accepts a cache_overlap_ns parameter. CACHE_READ gets array_ns = tDCBSYR1 with the remainder as xfer_ns. CACHE_PROG gets
  array_ns = tCBSY. Non-cache ops are unchanged.
  3. Time-bounded mid-pipeline state (cmd_engine.c) — After CACHE_READ, the die's remaining_ns is set to tDCBSYR1. After CACHE_PROG, remaining_ns is set to tCBSY. Status snapshots now report a finite overlap window.
  4. SM-56 timing assertion (test_media.c) — Added EAT drain between measurements and replaced (void) casts with TEST_ASSERT(cache_2_ns < plain_2_ns).

  Test results: 757/757 pass (132 common + 409 media + 61 HAL + 75 FTL + 44 controller + 36 shmem).

  Ready to commit and push when you give the word. Also noted your preference: local testing only going forward, no local GitHub Actions runner.

PR #83, opened 2026-04-19
[codex] add authoritative repo agent guide

AlvinYangZF:
## Summary

This PR adds a repository-local `AGENT.md` as the authoritative instruction file
for agents working in `hfsss-simulator`, and converts `CLAUDE.md` into a
compatibility pointer.

## Why

The repository already had meaningful project-specific guidance in
`CLAUDE.md`, but agent behavior is more reliable when there is a single source
of truth instead of multiple partially overlapping instruction files.

Adding `AGENT.md` makes the intended workflow explicit for tools that look for
that filename, while preserving compatibility for tools that still discover
`CLAUDE.md` first.

## What Changed

- Added `AGENT.md` as the primary guide for repository scope, operating
  principles, communication rules, build and test workflow, review expectations,
  and documentation conventions.
- Preserved the English-only output rule for agent messages, PR comments, and
  newly added code comments.
- Converted `CLAUDE.md` into a short pointer that directs agents to `AGENT.md`
  instead of maintaining two competing instruction sets.

## Validation

This is a documentation-only change. I verified the branch diff is limited to
`AGENT.md` and `CLAUDE.md`, and `git diff --check origin/master...HEAD` returns
cleanly.

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -1,0 +1,142 @@
+# AGENT.md
+
+This file is the authoritative instruction set for agents working in the
+`hfsss-simulator` repository. If another agent guidance file exists, follow this
+file unless the user gives a more specific instruction in the current session.
+
+## Repository Scope
+
+HFSSS is a full-stack SSD simulator written in C. The codebase spans:
+
+- NVMe and PCIe user-space interfaces
+- SSD controller logic
+- FTL, mapping, GC, wear leveling, ECC, and recovery
+- HAL and NAND media simulation
+- Stress, system, and unit tests
+- Investigation tooling under `scripts/`, `tools/`, and `docs/superpowers/`
+
+The layering matters. Keep changes localized and preserve interface boundaries
+between `include/`, `src/`, `tests/`, and `docs/`.
+
+## Operating Principles
+
+- Explain the intended approach before implementing meaningful changes.
+- If a request is ambiguous, high-risk, or high-impact, clarify the constraint
+  or approval boundary before writing code.
+- In plan-only workflows, write proposals only and do not write code.
+- Prefer spec-driven work for multi-step, high-risk, or architecture-affecting
+  changes. For small, well-scoped fixes, proceed directly once the scope is
+  clear.
+- Keep changes independently verifiable. Break large work into loosely coupled
+  slices when possible.
+- If the same manual process repeats several times, turn it into a script,
+  helper, or explicit repository rule.
+- Separate implementation from review. Finish the change first, then review it
+  with fresh eyes.
+- When fixing bugs, reproduce first when practical, then fix, then verify.
+
+## Language And Communication
+
+- Use English only in agent outputs for this repository.
+- Use English only in PR comments and review comments.
+- Use English only in newly added code comments, documentation edits, and test
+  descriptions unless the user explicitly requests otherwise.
+- Do not use AI tool names in code comments, commit messages, or PR bodies.
+
+## Coding Style
+
+- Use C11 and follow Google-style C/C++ formatting conventions where they fit
+  this codebase.
+- Keep comments conceptual and technical. Do not write process-oriented comments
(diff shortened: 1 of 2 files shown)


AlvinYangZF (commented):
Review of this PR, findings ordered by severity.

## HIGH — material concerns

**H1. Claude Code reads `CLAUDE.md`, not `AGENT.md` — the pointer may silently drop context.**
Claude Code injects `CLAUDE.md` contents into the agent's system prompt directly; it does not auto-follow a "see other file" pointer. After this PR, a live Claude Code session will see only a 12-line pointer and lose the richer rules in `AGENT.md` (multi-thread FTL/NBD risk areas, trace seam discipline, English-only, etc.). Two ways to keep parity:
- Keep `CLAUDE.md` as a full mirror (symlink or generated from `AGENT.md`), or
- Inline the most load-bearing rules into `CLAUDE.md` and keep `AGENT.md` as the canonical human-facing file.

As written, tool A knows the rules and tool B does not — the exact "split authority" the PR says it is preventing.

**H2. Lost content vs. the file it replaces.** Items in the old `CLAUDE.md` that are not carried over:
- Project structure directory tree (fast orientation for a fresh agent).
- Test-health baseline ("1,750+ assertions").
- Explicit slash-command guardrails: `Never use /init`, `/loop`, `/simplify`, `/batch` — these are project-specific tool controls, not generic fluff.
- Forbidden progress-marker term list (`FIXED, Step, Week, Section, Phase, AC-x`) — the new file softens this to "progress markers", which loses the concrete prohibition a reviewer/linter would enforce.

These should be ported over before merge, not silently dropped.

**H3. Build-variant coverage is incomplete.** `AGENT.md` only documents `TRACE=1`. The Makefile also exposes `COV=1` (`build-cov`), `ASAN=1` (`build-asan`), `UBSAN=1` (`build-ubsan`), `TSAN=1` (`build-tsan`). These are exactly the variants reviewers reach for when triaging races / memory / coverage gaps. Omitting them encourages agents to reuse `build/` artifacts across flags and hit the stale-object trap the Makefile comment already warns about. Recommend a short `flag → BUILD_DIR` matrix.

## MEDIUM — accuracy

**M1. "C11" is imprecise.** The Makefile uses `-std=gnu11`. Should say "GNU C11" or "C11 with GNU extensions" so agents do not reject valid GNU-only constructs (typeof, statement expressions, designated initializers with gaps) as non-portable.

**M2. Undefined acronym `TAA`.** Referenced in High-Risk Areas. Real subsystem (`include/ftl/taa.h`), but a fresh agent should not have to grep to resolve an acronym in an authoritative doc. Expand once, e.g. "TAA (throttle-aware allocator in `include/ftl/taa.h`)".

**M3. "Language And Communication" repeats itself.** Three consecutive bullets each start "Use English only in…". Consolidate into one bullet with a sub-list — easier to scan, harder to violate partially.

## LOW — polish

**L1. The `CLAUDE.md` compatibility stub still has 12 lines with headers and a bulleted list.** If it is a pointer, make it a pointer: two lines, one link. The current shape encourages future drift back into partial duplication.

**L2. "Separate implementation from review" appears twice in slightly different wording** (Operating Principles + implied in Review Expectations). One statement is enough.

**L3. The rule "Do not claim an end-to-end diagnostic pipeline works unless a real artifact was produced and consumed" is valuable but abstract.** A one-line reference to the incident this guardrail came from (e.g. the fio-014/015 segmented bisection plan under `docs/superpowers/plans/`) gives future readers the story without bloating the doc.

## What I would merge

Happy with: the consolidation intent, the new High-Risk Areas and Tooling/Diagnostic sections, explicit review-output expectations, and the mixed-scope PR rule.

Blocking before merge: H1 (agent parity) and H2 (lost content). H3 and the M-tier accuracy items are cheap to fix in the same PR and worth doing now rather than as a follow-up.

PR #82, opened 2026-04-19
feat: NAND Phase 7 — integration + profile matrix + long-form sweep (T1/T2/T4/T6)

AlvinYangZF:
## Scope

Phase 7 of the NAND command-coverage design (spec `docs/superpowers/specs/2026-04-10-nand-media-command-coverage-design.md` §11 Phase 7 + §12). Closes the validation-expansion gap left by Phase 1–6 (PRs #73/#74/#75/#76/#77/#79).

This PR lands 4 of the 6 planned Phase 7 tasks. **T3 (heavy combinations)** and **T5 (profile divergence + residual V3/V5/V6 + IS-09)** will land as follow-up PRs on this same branch or stacked PRs.

## Commits

| # | Commit | Task | Type |
|---|---|---|---|
| 1 | `b83a604` | **T1** basic integration (IS-05/06/08) | test-only |
| 2 | `db8d5a3` | **T2** mid integration (IS-02/03) + **DEAD abort-pattern runtime** | runtime + test |
| 3 | `c739bd5` | **T4** profile matrix V1/V2/V4 | test-only |
| 4 | `41e2c44` | **T6** long-form 4-profile sweep | test-only |

## Runtime change (T2 only)

Reset-abort now stamps the target page(s) of an interrupted PROG/CACHE_PROG/ERASE with the ` DE AD DE AD ...` pattern and marks them `PAGE_INVALID`. Previously they kept whatever partial bytes the aborted commit hook wrote (or stayed 0xFF), making an aborted page indistinguishable from a clean one. This is strictly additive — pre-existing tests don't hit the new path because they don't reset out of a non-IDLE state.

Touched files:
- \`include/media/nand.h\` — constants + \`page_size/spare_size\` cached on \`struct nand_device\`
- \`src/media/nand.c\` — \`nand_stamp_abort_pattern()\` helper + init wiring
- \`src/media/cmd_engine.c\` — stamp inside \`nand_cmd_engine_submit_reset\` under die_lock, before \`cmd_state_init\`

## New test binaries (480 Phase 7 assertions)

| Binary | Assertions | Scenarios |
|---|---|---|
| \`test_cmd_integration_basic\` | 78 | IS-05 init-time profile selection; IS-06 status/identity legality across busy states; IS-08 illegal suspend/resume rejection |
| \`test_cmd_integration_midcomplex\` | 57 | IS-02 non-conflicting reads during suspend; IS-03a reset-abort on PROG; IS-03b reset-abort on multi-plane ERASE |
[...]

PR #84, opened 2026-04-19
feat: FTL consumes NAND profile via HAL bridge

AlvinYangZF:
First **consumer-layer** PR for the Phase 7 profile divergence. Stacked on #82 (`feat/nand-cmd-engine-phase7`) because the query surface added here is only meaningful when ONFI and Toggle profiles actually diverge.

## Why

Phase 7 landed the divergence (supported_ops_bitmap + mp_rules.max_planes_per_cmd) entirely inside `src/media`. FTL and upper layers today do not know a profile exists — they reach into `hal_nand_read/program/erase_sync` and let the engine reject if something is unsupported, which surfaces as a generic \`HFSSS_ERR_INVAL\` without actionable context. This PR threads the active profile pointer through HAL → FTL so upper layers can preflight op support and size their plane batches without bouncing off a runtime rejection.

## Scope (minimal, thread-only)

**No behavioral change to the FTL dispatch path.** FTL still issues READ/PROG/ERASE only, all in the core bitmap. This is the wiring; follow-ups can widen the write path's plane batching using \`ftl_preferred_plane_batch\`, and guard future suspend/cache code paths with \`ftl_supports_op\`.

## API additions

| Function | Returns | NULL-safe behavior |
|---|---|---|
| \`hal_get_profile(hal_ctx)\` | profile pointer / NULL | NULL if HAL has no NAND, media uninitialized, or no profile |
| \`ftl_get_profile(ftl_ctx)\` | profile pointer / NULL | NULL ctx → NULL |
| \`ftl_preferred_plane_batch(ftl_ctx)\` | \`u8\` | NULL/no-profile → 1 (sequential default) |
| \`ftl_supports_op(ftl_ctx, op)\` | \`bool\` | NULL/no-profile → true (legacy permissive) |

Profile is resolved **once at \`ftl_init\`** and never mutated — a profile switch would require tearing down the FTL stack (mapping, CWB, block manager are keyed off profile-derived geometry and capability in later phases).

## Layering

```
upper layer (NBD / controller / tests)
       ↓ ftl_get_profile / ftl_preferred_plane_batch / ftl_supports_op
   FTL (ctx->profile cached at init)
       ↓ hal_get_profile(hal_ctx) at init only
   HAL (hal_nand_dev->media_ctx → struct media_ctx * cast)
       ↓
   media (profile resolved at media_init)
```

HAL keeps zero mutable profile state — it's a pass-through. The void-pointer cast on \`media_ctx\` is the same pattern HAL already uses for \`read/program/erase\` (\`src/hal/hal_nand.c:51,73,95,...\`).

## Tests (58 assertions, \`tests/test_ftl_profile.c\`)

| Test | Coverage |
|---|---|
| \`test_profile_thread\` | All 4 profiles: \`hal_get_profile == ftl_get_profile\`, id matches config, interface_family matches expectation. NULL inputs → NULL. |
| \`test_preferred_plane_batch\` | ONFI = 4, Toggle = 2, NULL ctx = 1. |
| \`test_supports_op\` | ONFI accepts \`READ_STATUS_ENHANCED\`; Toggle rejects it. Core ops (READ/PROG/ERASE/RESET/READ_STATUS) supported everywhere. NULL ctx → true. |
| \`test_write_read_parity\` | \`ftl_write\`/\`ftl_read\` round-trip on 4 profiles × 2 patterns. Verifies no regression in the FTL dispatch path. |

## Regression

\`make test\` full suite green with no Phase 7 assertion lost:
- \`test_media\` 473/473
- Phase 7 suites 78+57+83+210+196 (unchanged)
- \`test_ftl_profile\` **58/58** (new)
- All other unit + system tests unchanged

## Stacked PR notes

- Base: \`feat/nand-cmd-engine-phase7\` (PR #82)
- When #82 merges to master, this PR's base auto-rebases.
- Review can proceed independently of #82 since this PR adds only new public APIs and one init-time resolution step.

## Test plan
- [x] \`make all\` clean
- [x] \`make test\` full green
- [x] \`build/bin/test_ftl_profile\` standalone pass
- [x] NULL-input paths do not crash
- [x] No regression in FTL write/read


AlvinYangZF:
Review findings on the current head `18e6b8d`.

1. Medium: `ftl_preferred_plane_batch()` currently returns the profile cap, not the cap that is actually usable on the configured device geometry. In `src/ftl/ftl.c`, it returns `ctx->profile->mp_rules.max_planes_per_cmd` directly. But the new test stack in `tests/test_ftl_profile.c` is configured with only `planes_per_die = 2`, and still asserts `4` for ONFI. Since this API is described as something upper layers can use to size plane batches, returning `4` on a 2-plane device is a misleading contract. I think this helper should clamp to `min(profile->mp_rules.max_planes_per_cmd, ctx->config.planes_per_die)` (with the same legacy fallback to `1`).

2. Medium: the new validation is not wired into the repo’s coverage workflow. `TEST_FTL_PROFILE` is added to `all` / `test`, but it is missing from `COVERAGE_UT_BINS` in `Makefile`, so `make coverage-ut` / `make coverage` can still pass without building or running the only test that exercises this new HAL -> FTL profile bridge.

This review is based on static inspection of PR `#84` against its actual base branch `feat/nand-cmd-engine-phase7`; I did not run the branch locally in this pass.

AlvinYangZF:
Rebased onto master after PR #82 merged (merge commit `a525557`).

## Changes since previous review

- Base switched: `feat/nand-cmd-engine-phase7` → `master`
- One Makefile conflict resolved (both `TEST_RESET_ABORT_RACE` from round-1 fixes and `TEST_FTL_PROFILE` from this PR kept — disjoint intent)
- Single commit rebased clean: `e1d623d feat: FTL consumes NAND profile via HAL bridge` (was `18e6b8d`)

## Verification post-rebase

- `make all` clean
- `test_ftl_profile`: 58/58 pass
- Full `make test` suite: zero failing suites across all binaries (Phase 7 suites + round-1 fixes + legacy all green)

No functional changes from the pre-rebase version — this is purely a base update. Ready for review.

PR #85, opened 2026-04-19
feat: expose NAND profile selection via hfsss-nbd-server -P flag

AlvinYangZF:
## Summary

- Propagates the existing profile selection fields (`profile_id`,
  `profile_explicit`) through `sssim_config` into `media_config`, closing
  the last gap that prevented selecting a non-default NAND profile from
  the top of the user-space stack.
- Adds `nand_profile_id_from_name()` so CLIs and config parsers can map
  stable short names (`onfi-tlc`, `onfi-qlc`, `toggle-tlc`, `toggle-qlc`)
  to enum ids without depending on integer ordering.
- Adds `-P <name>` to `hfsss-nbd-server` and updates the startup banner
  to print the selected profile and whether it was explicit.
- Threads `HFSSS_NBD_PROFILE` through the qemu-blackbox harness so cases
  can drive profile selection without case-script changes.

Back-compat contract: zero-initialized callers observe no behavior
change — the profile is still derived from `nand_type` via
`nand_profile_get_default_for_type`.

## Why

After the Phase 6/7 NAND work landed, all four generic profiles
(ONFI TLC/QLC, Toggle TLC/QLC) exist in the simulator core but the only
way to reach them was to construct a `media_config` by hand. The NBD
server — which is the real host-side entry point for QEMU E2E
validation — hard-coded the default. Any profile-aware testing through
the harness therefore had to touch source to switch profiles, which
undermines the whole point of having a profile abstraction.

## Validation

- **Unit test (new):** `tests/test_sssim_profile.c`, 14 assertions
  covering default-TLC back-compat, QLC-type back-compat, explicit
  Toggle-TLC / Toggle-QLC override, short-name lookup happy path, and
  unknown / NULL failure modes.
- **Full regression:** `make test` → 2752 PASS, 0 FAIL (new test
  integrated into the standard runner).
- **End-to-end (qemu-blackbox):** `HFSSS_NBD_PROFILE=toggle-tlc
  run_qemu_blackbox_tests.sh --mode mt` running
  `fio/020_pr_smoke_seqwrite_verify` + `nvme/001_nvme_cli_smoke`.
  - Both cases PASS.
  - fio crc32c verify: 8192 IOs, `error=0`.
  - NBD startup banner: `Profile:  generic_toggle_enterprise_tlc (explicit)`.
  - Unknown profile rejected with a clear error message.

## Known limitation

The NVMe Identify `mn` field visible in the guest still shows the QEMU
emulated controller, not the NAND profile identity. That is expected
for the current QEMU → NBD path, where QEMU owns the NVMe controller
surface. Surfacing the NAND persona through a vendor log page or a
top-level controller-identity override is tracked as follow-on work and
not included here.

## Test plan

- [x] `make test` clean across full repo
- [x] `make build/bin/test_sssim_profile && ./build/bin/test_sssim_profile` → 14/14 PASS
- [x] `build/bin/hfsss-nbd-server -h` shows the new `-P` entry
- [x] `build/bin/hfsss-nbd-server -P bogus` returns clear error
- [x] qemu-blackbox smoke (fio verify + nvme-cli) PASS under
      `HFSSS_NBD_PROFILE=toggle-tlc`


AlvinYangZF:
Review findings on the current head `94ef0d0`.

1. Medium: `src/vhost/hfsss_nbd_server.c:772` does not actually print the resolved active profile; it prints a pre-init guess derived from the local `profile_id` variable, and it prints the internal `id_string` rather than the stable CLI alias this PR introduces. In the default path, `profile_id` is hard-coded to `NAND_PROFILE_GENERIC_ONFI_TLC` and `profile_explicit=false`, so the banner is describing the requested/assumed profile, not the profile that the initialized stack actually resolved. Today that happens to line up because the server also defaults `nand_type` to TLC, but the implementation does not match the stated contract of “print the selected profile,” and the log output is not round-trippable back into `-P` because it emits `generic_toggle_enterprise_tlc` instead of `toggle-tlc`.

2. Medium: `Makefile:189` still does not add `test_sssim_profile` to `COVERAGE_UT_BINS`. The new test is wired into `all` and `test`, but `make coverage-ut` / `make coverage` still will not build or run the only regression that exercises the new top-of-stack `sssim_config -> media_config` profile propagation path. Since the PR explicitly adds a new validation surface, leaving it outside the repo’s coverage workflow is a verification gap.

This review is based on static inspection of PR `#85` at head `94ef0d0`; I did not run the branch locally in this pass.

AlvinYangZF:
Review findings addressed in commit `2e15cf8`.

**M1 — banner now prints the actually-resolved profile, with a round-trippable name.**
- Moved the `Profile:` line to *after* `nvme_uspace_dev_init` and resolved it from `dev.sssim.media.profile`, so the log describes the stack state, not the pre-init request.
- Added `nand_profile_name()` as the inverse of `nand_profile_id_from_name`. The banner now emits `toggle-tlc` rather than `generic_toggle_enterprise_tlc`, which is directly round-trippable back into `-P`.
- New unit assertions cover both the inverse lookup and a two-way round-trip (`id_from_name(name_for_id(x)) == x`), guarding against table drift between the two helpers.

Banner under `HFSSS_NBD_PROFILE=toggle-tlc` now reads:
```
HFSSS NBD Server
...
Initializing simulator...
Profile:  toggle-tlc (explicit)
```

**M2 — `test_sssim_profile` is now in `COVERAGE_UT_BINS`.**
While making the change I ran `scripts/coverage/check_target_set.sh` and it also surfaced two pre-existing gaps (`test_ftl_profile`, `test_trace`) that were dropped by prior merges. Plugged all three in the same edit so the self-check goes green again rather than continuing to mask the drift.

**Validation on head `2e15cf8`:**
- `./build/bin/test_sssim_profile`: **21/21 PASS** (14 original + 7 new reverse/round-trip).
- `make test`: **2759 PASS, 0 FAIL**.
- `bash scripts/coverage/check_target_set.sh`: **OK**.
- `qemu-blackbox` smoke under `HFSSS_NBD_PROFILE=toggle-tlc`: fio crc32c seqwrite verify + nvme-cli smoke both PASS; banner verified in `nbd-server.log`.

AlvinYangZF:
TSAN findings from my Step 4 audit have been fixed on head \`c69b0c5\`.

### Root causes

**Race A — \`mutex_unlock\` unlock_count write (5–12 warnings per test).** \`mtx->unlock_count++\` ran *after* \`pthread_mutex_unlock\`, so concurrent acquirers could see the counter mutate outside the lock. The fix is one line: increment before the pthread unlock, matching the read-side guard in \`mutex_stats\`.

**Race B — \`eat_ctx\` compare-then-maybe-write (10 warnings + the \`remaining_ns > 0 while suspended\` failure).** \`eat_update\` does \`if (new_eat > cur) cur = new_eat\` on four array slots (channel/chip/die/plane) with no synchronization. Two writers observing the same stale \`cur\` can lose the later, larger update — violating the monotonic-max invariant that \`engine_wait_until_available_mask\` depends on. Added a recursive mutex inside \`eat_ctx\` and serialized every public \`eat_*\` helper through it.

### Why the systest_phase7 assertion was failing

Contention on unlocked EAT slots was delaying the suspend handler under TSAN's slowdown. That occasionally let the full tPROG elapse before \`engine_run_array_busy\` processed the suspend request, zeroing \`remaining_ns\`. With the EAT race closed the timing margin is preserved and the assertion holds.

### Validation on head \`c69b0c5\`

| Target | Before | After |
|---|---|---|
| \`make test\` (default) | 2759 PASS / 0 FAIL | 2759 PASS / 0 FAIL |
| \`test_reset_abort_race\` (TSAN) | 30/30, 5 warnings | **30/30, 0 warnings** |
| \`test_cmd_integration_midcomplex\` (TSAN) | 64/64, 12 warnings | **64/64, 0 warnings** |
| \`systest_phase7_integration\` (TSAN) | 195/196, 10 warnings, 1 FAIL | **196/196, 0 warnings** |
| qemu-blackbox smoke (toggle-tlc) | PASS | PASS |

### Diff

3 files, +48/−8: \`src/common/mutex.c\`, \`include/media/eat.h\`, \`src/media/eat.c\`. No API changes. Recursive mutex so internal helpers can call each other without deadlock (e.g. \`eat_get_max\` → \`eat_get_for_chip\`).

PR #86, opened 2026-04-19
fix(nvme): restore Identify Namespace data structure to 4096 bytes

AlvinYangZF:
## Summary

- `struct nvme_identify_ns` was 4108 bytes (NVMe spec requires 4096). `rsvd1[40]` retained the pre-1.4 size after ANAGRPID/NSATTR/NVMSETID/ENDGID were added in-place, double-counting 12 bytes.
- `nvme_build_identify_ns` issues `memset(id, 0, sizeof(*id))` = 4108 bytes into 4096-byte caller buffers (e.g. `calloc(1, 4096)` in `nvme_uspace_exercise_admin_path`), writing 12 bytes past the heap region.
- Fix: shrink `rsvd1` to 28 bytes, restoring NVMe 1.4 field offsets (ANAGRPID at 0x5C, NGUID at 0x68, EUI64 at 0x78, LBAF at 0x80, VS at 0x180). Added `_Static_assert(sizeof == 4096)` as a compile-time regression guard.

## Root Cause Evidence

```
==54712==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 4108 at 0x621a00060100
    #1 nvme_ctrl_process_identify nvme.c:399
    #2 nvme_uspace_dispatch_admin_cmd nvme_uspace.c:711
    #3 nvme_uspace_exercise_admin_path nvme_uspace.c:923
0 bytes after 4096-byte region
allocated by calloc at nvme_uspace.c:902
```

## Test Plan

- [x] Static assert fails compilation before fix (proves RED): `4108 == 4096` → compile error
- [x] Static assert passes after fix (proves GREEN): `sizeof == 4096` holds
- [x] ASAN build: nbd-server startup clean, Identify Controller + Identify Namespace both `rc=0`, server reaches "Listening" state without overflow report
- [x] `make test` regression: 0 failures across all test tallies
- [x] No callers reference the shifted field offsets in generated data — only `lbaf[0]` is populated (checked by grep)

### DIFF
--- a/include/pcie/nvme.h
+++ b/include/pcie/nvme.h
@@ -402,7 +402,12 @@ struct nvme_identify_ns {
     u16 nabspf;       /* 0x2C: Namespace Atomic Boundary Size Power Fail */
     u16 noiob;        /* 0x2E: Namespace Optimal I/O Boundary */
     u64 nvmcap[2];    /* 0x30: NVM Capacity */
-    u8  rsvd1[28];    /* 0x40: Reserved */
+    u16 npwg;         /* 0x40: Namespace Preferred Write Granularity */
+    u16 npwa;         /* 0x42: Namespace Preferred Write Alignment */
+    u16 npdg;         /* 0x44: Namespace Preferred Deallocate Granularity */
+    u16 npda;         /* 0x46: Namespace Preferred Deallocate Alignment */
+    u16 nows;         /* 0x48: Namespace Optimal Write Size */
+    u8  rsvd1[18];    /* 0x4A: Reserved */
     u8  anagrpid[4];  /* 0x5C: ANA Group Identifier */
     u8  rsvd2[3];     /* 0x60: Reserved */
     u8  attr;         /* 0x63: Namespace Attributes */
--- a/include/pcie/nvme.h
+++ b/include/pcie/nvme.h
@@ -402,23 +402,26 @@ struct nvme_identify_ns {
     u16 nabspf;       /* 0x2C: Namespace Atomic Boundary Size Power Fail */
     u16 noiob;        /* 0x2E: Namespace Optimal I/O Boundary */
     u64 nvmcap[2];    /* 0x30: NVM Capacity */
-    u8  rsvd1[40];    /* 0x40: Reserved */
-    u8  anagrpid[4];  /* 0x68: ANA Group Identifier */
-    u8  rsvd2[3];     /* 0x6C: Reserved */
-    u8  attr;         /* 0x6F: Namespace Attributes */
-    u16 nvmsetid;     /* 0x70: NVM Set Identifier */
-    u16 endgid;       /* 0x72: Endurance Group Identifier */
-    u8  nguid[16];    /* 0x74: Namespace Globally Unique Identifier */
-    u64 eui64;        /* 0x84: IEEE Extended Unique Identifier */
+    u8  rsvd1[28];    /* 0x40: Reserved */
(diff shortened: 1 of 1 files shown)


AlvinYangZF:
### Independent review summary

Ran an independent second-opinion pass (via the codex-rescue sub-agent) against head `3ab1868`. Overall verdict: **approve with one nit**, addressed in follow-up commit `41c152c`.

#### 1. Spec correctness — match
Computed offsets from the fixed struct vs NVMe 1.4 Figure 245:

| Field | Fixed struct offset | NVMe 1.4 | Verdict |
|---|---:|---:|---|
| `rsvd1` span (fixed) | `0x40..0x5b` | `0x40..0x5b` | byte span matches |
| `anagrpid[4]` | `0x5c` | `0x5c` | ✓ |
| `rsvd2[3]` | `0x60` | `0x60` | ✓ |
| `attr` (NSATTR) | `0x63` | `0x63` | ✓ |
| `nvmsetid` | `0x64` | `0x64` | ✓ |
| `endgid` | `0x66` | `0x66` | ✓ |
| `nguid[16]` | `0x68` | `0x68` | ✓ |
| `eui64` | `0x78` | `0x78` | ✓ |
| `lbaf[16]` | `0x80` | `0x80` | ✓ |
| `rsvd3[192]` | `0xc0` | `0xc0` | ✓ |
| `vs[3712]` | `0x180` | `0x180` | ✓ |
| total | `0x1000` | `0x1000` | ✓ |

Nit: `rsvd1[28]` spans the NPWG/NPWA/NPDG/NPDA/NOWS region plus 18 reserved bytes per NVMe 1.4 Figure 245, so labelling all 28 bytes "reserved" understates what the spec defines.

**Addressed in `41c152c`**: split into `u16 npwg/npwa/npdg/npda/nows` (10 bytes) + `u8 rsvd1[18]` (18 bytes). Byte span unchanged, static assert still passes, regression clean.

#### 2. Hidden callers — none
Grepped `src/`, `tests/`, and `tools/` for any code reading/writing the old raw offsets (`0x74`, `0x84`, `0x88`) or using `offsetof(struct nvme_identify_ns, ...)`. No matches. Current consumers only touch typed fields — notably `lbaf[0]` in `src/pcie/nvme.c:377-379`, `src/pcie/nvme_uspace.c:199-210`, and several tests.

#### 3. Static assert as regression guard — good fence
`u8/u16/u32/u64` come from fixed-width `stdint.h` typedefs in `include/common/common.h:14-17` and the struct is `__attribute__((packed))` at `include/pcie/nvme.h:421`. On GCC/Clang (the only toolchains this repo already targets via other GNU-only features), `sizeof == 4096` is deterministic. Requires C11 `_Static_assert`, which the repo already assumes.

#### 4. Shrink vs. caller-side `sizeof()` — shrink was correct
The alternative (bump all callers to `sizeof(struct nvme_identify_ns)`) would have masked the host overflow without fixing the wire image. With the pre-fix layout, fields after `0x5b` were shifted by 12 bytes, so a guest or parser reading the spec-defined 4096-byte Identify payload would have found ANAGRPID/NSATTR/NVMSETID/ENDGID/NGUID/EUI64/LBAF in the wrong places. The overflow was not host-only — it was wire-format corruption for any standards-compliant 4096-byte consumer.

#### Suggested follow-up (separate PR)
Introduce an `NVME_IDENTIFY_DATA_SIZE 4096` constant and replace the literal `4096` in `src/pcie/nvme.c:386` and the allocation/dispatch sites in `src/pcie/nvme_uspace.c`. Keeps the static assert as the invariant source.

PR #88, opened 2026-04-19
docs: sync REQ coverage, roadmap, traceability matrix to current code

AlvinYangZF:
## Summary

The three REQ tracking docs (`REQUIREMENT_COVERAGE.md`, `IMPLEMENTATION_ROADMAP.md`, `TRACEABILITY_MATRIX.md`) had drifted significantly from reality — the coverage doc listed many features as ❌ / ⚠️ / 🔧 despite having landed implementations + passing tests in the source tree, and the roadmap contradicted itself (snapshot table claimed Phases 4-7 ✅ while the per-phase sections were still marked "In Progress" / "Queued").

Audit + sync:

- **Aggregate change**: Core 50.0% → **66.7%** implemented (core ⚠️ included: 83.3%). Grand total 38.8% → **68.5%** (86.5% counting partials). 122 ✅ / 31 ⚠️ / 24 ❌ / 1 🔧 across the 178 REQs.
- **No code changes.** Evidence is file-paths-and-tests referenced in each row's note column.

## Per-module shifts

| Module | Before | After |
|--------|-------:|------:|
| PCIe/NVMe | 50.0% | 54.5% |
| Controller Thread | 60.0% | 80.0% |
| Media Threads | 70.0% | 75.0% |
| HAL | 91.7% | 75.0% (AER flipped to ⚠️) |
| Common Services | 37.5% | **75.0%** |
| FTL | 59.1% | **81.8%** |
| Product Interfaces | 0% | **50.0%** |
| Fault Injection | 0% | **66.7%** |
| Enterprise UPLP | 0% | **100%** |
| Enterprise Multi-NS | 0% | **100%** |
| Enterprise Security | 0% | **85.7%** |
| Enterprise Thermal/Telemetry | 0% | **75.0%** |

## Notable ❌ → ✅ flips (with impl + test binding)

- REQ-018 DSM/Trim — `nvme_uspace.c` + `test_dsm.c`
- REQ-053..056 full NOR (7-partition, mmap persistence) — `media/nor_flash.c` + `test_nor_flash.c`
- REQ-078..081 boot / dual-slot firmware / power mgmt — `common/boot.c` + `test_boot.c` + `test_power_cycle.c`
- REQ-082..084 OOB JSON-RPC + SMART — `common/oob.c` + `test_oob.c`
- REQ-091 trace ring (compile-gated) — `common/trace.c` + `test_trace.c`
[...]

PR #89, opened 2026-04-19
feat: Tier-A REQ closure (5 flips from ⚠️/❌ → ✅)

AlvinYangZF:
## Summary

Closes the four Tier-A items from the post-PR-#88 residual list, with a genuine bug fix tucked inside REQ-157.

| REQ | Before | After | Commit | Evidence |
|---|---|---|---|---|
| **REQ-164** Secure boot verify | ⚠️ | ✅ | `ce738c8` | `secure_boot_verify()` invoked during `BOOT_PHASE_1_POST`; tampered image / wrong magic → `HFSSS_ERR_AUTH` abort |
| **REQ-163** Sanitize action modes | ⚠️ | ✅ | `4f36140` | All four SANACT modes dispatched (EXIT_FAILURE / BLOCK_ERASE / OVERWRITE / CRYPTO_ERASE); OVERWRITE performs explicit zero-fill per LBA |
| **REQ-115** NVMe Error Log Page | ⚠️ | ✅ | `c677eab` | `Get Log Page LID=0x01` now returns a 64-entry ring populated via `nvme_uspace_report_error()`; status_field carries SCT/SC per spec |
| **REQ-158** E2E PI error reporting | ⚠️ | ✅ | `c677eab` | PI errors map to NVMe status (SCT=2, SC=0x81..0x83) and flow into the Error Log Page |
| **REQ-157** PI propagation through GC | ⚠️ | ✅ | `cffadd3` | **Genuine bug fixed**: `media_inject_bit_errors` was silently dropping the spare buffer on every NAND read, so GC migrates lost T10 PI |

## Bug fix — REQ-157 was not just a wiring gap

`src/media/media.c::media_inject_bit_errors()` copied `data` back to the caller but cast `(void)spare`. All HAL `hal_nand_read_sync()` calls returned the data buffer correctly, but the caller's `spare` buffer was untouched. GC's migrate path (which already correctly passes `spare_buf` through `hal_nand_read_sync` → `hal_nand_program_sync`) was therefore dragging zeroed / sentinel bytes forward and every migrated page lost its PI tuple. `tests/test_t10_pi.c::test_pi_gc_preservation` now drives a real end-to-end scenario (media + HAL) that fails pre-fix and passes post-fix.

## Refactor prerequisite

`secure_boot_verify()` + `struct fw_signature` + `FW_SIG_MAGIC` were moved from `controller/security.{h,c}` into `common/boot.{h,c}`. Rationale: the boot sequence wasn't previously linking the controller library, and every test that currently links `libhfsss-common` would have needed `libhfsss-controller` as a transitive dependency just to satisfy `secure_boot_verify`'s symbol. Moving the verifier into common keeps boot.c standalone. `controller/security.h` now re-exports via `#include "common/boot.h"`, so test_security.c and other security callers compile unchanged.

## Coverage (before → after Tier-A)

| | Before | After |
|---|---:|---:|
| ✅ Implemented | 116 | **121** |
| ⚠️ Partial | 38 | 33 |
| ❌ Not Implemented | 23 | 23 |
| 🔧 Stub | 1 | 1 |
| **Fully** | 65.2% | **68.0%** |
| **At-least-partial** | 86.5% | 86.5% |

Per-module notable shifts:
- FTL: 18 → 19 ✅ (Error Log Page now populated)
- Enterprise T10 DIF/PI: 3 → **5 ✅** (group 100% complete)
- Enterprise Security: 3 → 4 ✅ (secure boot wired)

## Test Plan

- [x] `test_boot` — 53/53, including 4 new secure-boot cases (valid image accept, tampered reject, bad magic reject, back-compat when no image attached)
- [x] `test_nvme_uspace` — 53/53, including 11 new sanitize-action-modes cases + 14 new Error Log Page cases (empty, ordering, ring wrap, undersized-buffer)
- [x] `test_t10_pi` — 24/24, including the new end-to-end `test_pi_gc_preservation` (fails pre-fix, passes post-fix)
- [x] Full `make test` regression: 0 failures
- [x] Docs (`REQUIREMENT_COVERAGE.md`, `TRACEABILITY_MATRIX.md`) updated mechanically; Coverage Summary recomputed

## Commits

1. `ce738c8` feat(boot): invoke secure_boot_verify in POST phase (REQ-164)
2. `4f36140` feat(nvme): dispatch Sanitize action modes (REQ-163)
3. `c677eab` feat(nvme): populate Error Information Log Page (REQ-115, REQ-158)
4. `cffadd3` fix(media): preserve spare area across NAND read (REQ-157)
5. `76aadae` docs: sync REQ coverage for Tier-A closure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -407,9 +407,9 @@ $(TEST_NVME_USPACE): $(TEST_DIR)/test_nvme_uspace.c $(LIBHFSSS_PCIE) $(LIBHFSSS_
 	@echo "  CC      $@"
 	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-pcie -lhfsss-sssim -lhfsss-controller -lhfsss-ftl -lhfsss-hal -lhfsss-media -lhfsss-common -lm $(LDFLAGS)
 
-$(TEST_BOOT): $(TEST_DIR)/test_boot.c $(LIBHFSSS_COMMON)
+$(TEST_BOOT): $(TEST_DIR)/test_boot.c $(LIBHFSSS_COMMON) $(LIBHFSSS_CTRL)
 	@echo "  CC      $@"
-	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-common $(LDFLAGS)
+	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-controller -lhfsss-common $(LDFLAGS)
 
 $(TEST_NOR): $(TEST_DIR)/test_nor_flash.c $(SRC_DIR)/media/nor_flash.c $(LIBHFSSS_COMMON)
 	@echo "  CC      $@"
--- a/Makefile
+++ b/Makefile
@@ -407,9 +407,9 @@ $(TEST_NVME_USPACE): $(TEST_DIR)/test_nvme_uspace.c $(LIBHFSSS_PCIE) $(LIBHFSSS_
 	@echo "  CC      $@"
 	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-pcie -lhfsss-sssim -lhfsss-controller -lhfsss-ftl -lhfsss-hal -lhfsss-media -lhfsss-common -lm $(LDFLAGS)
 
-$(TEST_BOOT): $(TEST_DIR)/test_boot.c $(LIBHFSSS_COMMON) $(LIBHFSSS_CTRL)
+$(TEST_BOOT): $(TEST_DIR)/test_boot.c $(LIBHFSSS_COMMON)
 	@echo "  CC      $@"
-	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-controller -lhfsss-common $(LDFLAGS)
+	@$(CC) $(CFLAGS) $< -o $@ -L$(LIB_DIR) -lhfsss-common $(LDFLAGS)
 
 $(TEST_NOR): $(TEST_DIR)/test_nor_flash.c $(SRC_DIR)/media/nor_flash.c $(LIBHFSSS_COMMON)
 	@echo "  CC      $@"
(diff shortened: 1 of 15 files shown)


AlvinYangZF:
Review findings from a full pass over PR #89:

1. High: `src/common/boot.c:252-263` still makes secure boot opt-in instead of part of the normal boot path. `phase1_post()` only verifies when `fw_image/fw_sig` were pre-attached, and in this tree the only `boot_attach_fw_image()` call sites are the new unit tests (`tests/test_boot.c:399`, `:428`, `:453`). The new `test_secure_boot_skipped_when_no_attach()` (`tests/test_boot.c:462-472`) explicitly locks in that the default boot path bypasses verification entirely. So a real boot that never populates those pointers will still accept tampered firmware, which means REQ-164 is still overclaimed here.

2. High: `src/pcie/nvme_uspace.c:544-563` adds an Error Information Log ring, but `nvme_uspace_report_error()` still has no non-test callers. The only call sites in the tree are the manual injections in `tests/test_nvme_uspace.c:246-250` and `:274-275`. That means `Get Log Page LID=0x01` is still empty for real UCE / PI failures, so REQ-115 / REQ-158 are not actually wired through the production path yet.

3. Medium: `src/pcie/nvme_uspace.c:618-632` still hardcodes SMART `num_err_log_entries` to zero even though `include/pcie/nvme_uspace.h:38-44` now documents `error_log_count` as being exposed there. I reproed this locally against the PR checkout: after calling `nvme_uspace_report_error()`, `nvme_uspace_get_log_page(..., NVME_LID_SMART, ...)` still returns `num_err_log_entries=0`. So the SMART log and the Error Information log disagree from the host’s point of view.

4. Low: the summary tables are now internally inconsistent. `docs/REQUIREMENT_COVERAGE.md:44` says Enterprise: Security is `4 implemented / 3 partial / 57.1%`, while `docs/TRACEABILITY_MATRIX.md:288` says the same group is `5 implemented / 2 partial / 71.4%`.

I also built and ran the touched focused tests on the PR branch (`test_nvme_uspace`, `test_boot`, `test_t10_pi`); they pass, but the gaps above remain because the new tests mostly validate helper-level behavior rather than production-path integration.

PR #90, opened 2026-04-20
feat: Tier-B REQ closure (4 flips ⚠️ → ✅ on top of PR #89)

AlvinYangZF:
## Summary

Four Tier-B requirements closed on top of PR #89 (`feat/tier-a-req-closure`). Each REQ is an independent commit with its own tests and docs sync.

- **REQ-174 / REQ-175 / REQ-176** — NVMe Telemetry + Vendor Log Pages (0x07 / 0x08 / 0xC0) dispatched through `nvme_uspace_get_log_page` off the existing `src/common/telemetry.c` ring. Generation-number and data-available semantics match NVMe spec §5.14.1.7/8; vendor LID 0xC0 emits `struct nvme_vendor_log_counters`.
- **REQ-132** — `fault_registry` now consulted inside `media_nand_read/program/erase` via a new `media_attach_fault_registry` hook. Injected faults surface as `HFSSS_ERR_IO` at the media boundary.
- **REQ-165** — `key_table_save`/`load` rerouted to a new `NOR_PART_KEYS` partition with dual-copy slots (A at 0, B at 64 KB), monotonic generation numbers, and outer CRCs. Save writes slot B first then slot A; load picks the highest valid generation.
- **REQ-063** — AER pending-event / outstanding-CID pair-matching framework in `src/hal/hal_aer.c` (LLD_13 §5.1). Integrated into `nvme_uspace_dev`; `NVME_ADMIN_ASYNC_EVENT` routes through `hal_aer_submit_request`; controller-side callers can post events via `nvme_uspace_aer_post_event`.

## Coverage shift

- Grand Total: 121 ✅ → **127 ✅** (68.0% → **71.3%**).
- HAL: 9 ✅ → **10 ✅** (75.0% → 83.3%, REQ-063).
- Fault Injection: 1 ✅ → **2 ✅** (33.3% → 66.7%, REQ-132).
- Security: 5 ✅ → **6 ✅** (71.4% → 85.7%, REQ-165; only REQ-161 TCG Opal remains ⚠️).
- Thermal/Telemetry: 4 ✅ → **7 ✅** (50.0% → 87.5%, REQ-174/175/176).

## Test plan

- [x] `test_hal`: 109/109 (includes 5 new AER test functions, 46 assertions).
- [x] `test_nvme_uspace`: 128/128 (new telemetry log page + AER admin dispatch coverage, +61 assertions).
- [x] `test_media`: 489/489 (new fault_injection_nand_path, +16 assertions).
- [x] `test_security`: 71/71 (new NOR-backed key table with 6 scenarios including mid-save crash recovery, +31 assertions).
- [x] Full `make test`: **2992 PASS / 0 FAIL** (was 2923 pre-Tier-B).

## Base branch

This PR targets **`feat/tier-a-req-closure`** (PR #89) rather than master. Once PR #89 merges, this PR will rebase onto master with no conflicts.


AlvinYangZF:
Review findings from a local diff against base `feat/tier-a-req-closure` plus targeted test runs.

1. High: REQ-165 looks overclaimed. The generic persistence entry points are still file-backed in `src/controller/security.c:241` (`key_table_save`) and `src/controller/security.c:263` (`key_table_load`). The NOR path was added separately in `src/controller/security.c:372` / `:409`, but I only found test call sites for it, for example `tests/test_security.c:555`. So the PR body/docs say save/load was rerouted, but the runtime path was not.

2. Medium: The telemetry log-page plumbing is present in `src/pcie/nvme_uspace.c:760`, but I found no non-test producer for the backing ring. `telemetry_record()` in `src/common/telemetry.c:33` is only exercised from tests such as `tests/test_nvme_uspace.c:444`. Based on that, LIDs `0x07`, `0x08`, and `0xC0` will stay empty/default in normal runtime unless another patch wires real event sources.

3. Medium: `nvme_uspace_dev_init()` added `telemetry_lock`, `telemetry`, and `aer`, but the later failure legs do not unwind them. The new resources are initialized at `src/pcie/nvme_uspace.c:67` and `:82`, while the later error paths at `src/pcie/nvme_uspace.c:92-124` mostly only clean up `dev->lock` and older subsystems. Any mid-init failure now leaks newly added state and makes cleanup behavior inconsistent.

Advice:
- Route the existing key persistence API through NOR before marking REQ-165 complete, or keep the docs/PR text scoped to “NOR path added” rather than “save/load rerouted”.
- For telemetry/AER, add at least one real producer path in this PR or downgrade the docs to “framework landed, producers pending”.
- Rework `nvme_uspace_dev_init()` to use a single `goto fail` unwind path so every resource is released in reverse init order.

Verification:
- Built and ran `test_hal`, `test_nvme_uspace`, `test_media`, and `test_security` on the PR head.
- All four passed locally.

AlvinYangZF:
Follow-up review on the latest PR #90 head `c9ed7e8` (`fix(nvme,security,docs): address PR #90 self-review findings`).

The latest fixup does address most of the earlier review:

- `key_table_save/load` are now NOR-only canonical APIs in `include/controller/security.h` and `src/controller/security.c`; the old file-backed overloads are gone.
- `nvme_uspace_dev_init()` now uses a proper reverse-order unwind path in `src/pcie/nvme_uspace.c`, so the telemetry/AER init leak is fixed.
- The docs now explicitly keep `REQ-178` partial and add caveats for `REQ-174/175/176`, which is a much more accurate description of the current runtime state.

Residual findings on the current head:

1. Medium: The telemetry producer gap still remains. `telemetry_record()` is still only exercised from tests such as `tests/test_nvme_uspace.c:444`, `:521`, and `:573`; I did not find a non-test producer in `src`. The log-page dispatch itself is real in `src/pcie/nvme_uspace.c:762`, but in normal runtime LIDs `0x07`, `0x08`, and `0xC0` still stay effectively empty until Tier-C wiring lands. The docs now admit this explicitly, so this is a residual runtime gap rather than a docs/code mismatch.

2. Medium: REQ-165 is materially improved, but it still looks like an API-level fix rather than end-to-end runtime persistence. The canonical API is now NOR-backed in `src/controller/security.c:316` / `:353`, and tests use it in `tests/test_security.c:523`, but I still found no non-test caller. The live sanitize path in `src/pcie/nvme_uspace.c:491` still just trims and flushes without touching the key table, so I do not yet see a real startup/save/load flow.

3. Low: The PR body is stale versus the current head. It still reports `test_security: 71/71`, while the current head now runs `131/131` after the follow-up refactor.

Verification on current head `c9ed7e8`:
- `test_hal`: 109/109
- `test_nvme_uspace`: 128/128
- `test_media`: 489/489
- `test_security`: 131/131

Net: the follow-up commit fixed the earlier docs/API cleanup issues, and I do not see a new regression in that patch. The meaningful remaining gaps are still real telemetry producers and real end-to-end key persistence wiring.

PR #97, opened 2026-04-20
feat(nvme): wire per-NS QoS caps + SLA rollback + hot-reconfigure (REQ-148/149/150/151)

AlvinYangZF:
## Summary

The token-bucket + latency-monitor primitives in \`src/controller/qos_policy.c\` and \`src/controller/latency_monitor.c\` already existed but had no non-test callers. This PR wires the full QoS stack through the NVMe I/O surface in a single tight commit.

## What's new

- **REQ-148 / REQ-149** — \`nvme_uspace_read\` / \`nvme_uspace_write\` now refill + acquire tokens on every LBA-bearing command. Throttled commands return \`HFSSS_ERR_BUSY\`; dispatcher maps to CQE \`SC=NAMESPACE_NOT_READY\` (0x81) so hosts treat it as retryable.
- **REQ-151** — \`nvme_uspace_dev_set_qos_policy(nsid, policy)\` rebuilds token buckets in place. Next \`qos_acquire_tokens\` call sees the new caps — no traffic stop, no drain, no restart.
- **REQ-150** — \`nvme_uspace_dev_set_sla_rollback(nsid, target_us, trigger_count, cb, ctx)\` installs a P99 rollback. \`nvme_uspace_dev_check_sla\` folds trigger-count semantics on top of the existing \`lat_monitor_check_sla\` and resets consecutive_violations after fire so each window is evaluated independently.

## Coverage delta

Grand Total: **144 ✅ / 14 ⚠️ (80.9%)** — first crossing of the 80% milestone.

- Enterprise QoS Determinism: 2/5 → **6/1** (28.6% → 85.7%).
- Enterprise Subtotal: 35/5 → **39/1** (87.5% → 97.5%).
- Grand Total: 140/18 → **144/14** (78.7% → 80.9%).

## Test plan

- [x] \`test_nvme_uspace\`: 244 → **298** (+54 assertions: IOPS cap 13 + BW cap 12 + hot-reconfig 8 + SLA rollback 14, plus a dispatch CQE status assertion).
- [x] Full \`make test\`: **0 FAIL**.

## Key design choices

- QoS state lives **inside \`struct nvme_uspace_dev\`**, not as a sidecar registry. Callers don't need a second handle to wire the policy.
- Throttle returns at the API layer (\`HFSSS_ERR_BUSY\`) and the dispatcher does the NVMe-code mapping, matching the Opal / REQ-161 layering.
- SLA rollback \`trigger_count=0\` disables the callback without touching the monitor, so callers can hot-swap policies without re-arming latency tracking.

### DIFF
--- a/docs/REQUIREMENT_COVERAGE.md
+++ b/docs/REQUIREMENT_COVERAGE.md
@@ -39,13 +39,13 @@ This document analyzes the coverage of the 178 requirements from the Requirement
 | System Reliability | 4 | 3 | 0 | 1 | 0 | 75.0% | ↑ REQ-088 P99.9 latency anomaly detector |
 | **Core Subtotal** | **138** | **105** | **13** | **20** | **0** | **76.1%** (85.5% partial) | ↑ from 50.0% |
 | Enterprise: UPLP | 8 | 8 | 0 | 0 | 0 | 100% | ↑ implemented |
-| Enterprise: QoS Determinism | 7 | 2 | 5 | 0 | 0 | 28.6% (100% partial) | ↑ DWRR + partial wiring |
+| Enterprise: QoS Determinism | 7 | 6 | 1 | 0 | 0 | 85.7% (100% partial) | ↑ DWRR + per-NS IOPS/BW caps + SLA rollback + hot-reconfig |
 | Enterprise: T10 DIF/PI | 5 | 5 | 0 | 0 | 0 | 100% | ↑ Type 1/2/3 CRC-16 + GC-path PI propagation |
 | Enterprise: Security | 7 | 7 | 0 | 0 | 0 | 100% | ↑ AES-XTS sim, keys, crypto erase, sanitize action modes, secure-boot-verify, NOR-backed dual-copy UPLP-safe key table, TCG-Opal lock/unlock |
 | Enterprise: Multi-Namespace | 5 | 5 | 0 | 0 | 0 | 100% | ↑ implemented |
 | Enterprise: Thermal/Telemetry | 8 | 8 | 0 | 0 | 0 | 100% | ↑ throttle + SMART predict + NVMe Log Page 07h/08h/0xC0 dispatch + AER notifier helpers + REQ-178 runtime producer (smart_monitor) |
-| **Enterprise Subtotal** | **40** | **35** | **5** | **0** | **0** | **87.5%** (100% partial) | ↑ from 0% |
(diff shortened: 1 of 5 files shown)


AlvinYangZF:
All 7 Codex findings verified against the code and addressed in commit \`f5ce67e\`.

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | High | Hot-reconfigure raced live I/O — setter took no lock, read/write used keys_lock | New \`dev->qos_lock\` (leaf). Taken inside keys_lock on read/write around just the QoS access; taken in set_qos_policy / set_sla_rollback / check_sla / record_latency. All accessors agree. |
| 2 | High | Setter accepted phantom namespaces (nsid 2..32) despite Identify NN=1 | All four entry points now reject \`nsid != 1\` so the management API mirrors the I/O path |
| 3 | Medium | No spec-range enforcement on policy values | \`qos_policy_within_spec\` rejects sub-1K / super-2M IOPS, sub-50 / super-14000 MB/s BW (zero = unlimited stays legal) |
| 4 | Low | One-line rationale for SC=0x81 mapping | Expanded to cite NVMe Base Spec 2.0b §5 + DNR=0 retry semantics + fallback plan if real host interop shows otherwise |
| 5 | High | REQ-150 not wired into runtime I/O | \`nvme_uspace_dispatch_io_cmd\` stamps \`get_time_ns\` on entry + exit, calls \`nvme_uspace_dev_record_latency\` for LBA-bearing opcodes. Admin commands skip so they don't skew the histogram |
| 6 | Medium | \`check_sla\` callback-reentrancy race | Snapshot cb/ctx/p99 inside qos_lock; release lock; invoke callback outside. A callback that reinstalls the rollback no longer races the post-callback updates |
| 7 | Low | Missing tests for combined IOPS+BW, phantom nsid, reentrancy, dispatch wiring | +5 new tests: phantom nsid=2 rejection; IOPS range rejection; BW range rejection; combined IOPS+BW drain; reentrant-callback re-install; dispatch wire-in records >=2 samples |

### Verification

- \`test_nvme_uspace\`: 298 → **313** (+15 assertions).
- Full \`make test\`: **0 FAIL**.

### Knock-on adjustment

Raising the spec lower bound tripped two pre-existing tests that used \`iops_limit=100\` as a "tight cap" value. Updated them to \`iops_limit=1000\` (spec lower bound) and bumped iteration counts to 3000 so the bucket still drains under refill pressure.

Coverage status unchanged — REQ-148/149/150/151 stay ✅, Grand Total still 144/18 (80.9%).

PR #98, opened 2026-04-21
feat(perf,nbd): measured scalability + CPU/DRAM gates + nbd sampler wiring

AlvinYangZF:
## Summary

Two perf REQ flips plus a production wire-in for the process-wide resource sampler.

## What's new

- **REQ-122 — Scalability.** `perf_validation_run_all` now runs a **measured** scalability check: back-to-back `bench_run` invocations at nt=1 and nt=8 compute `iops(N) / (N * iops(1))`. A real regression in the worker path lowers the ratio and fails the 70% gate. The Amdahl model stays as a design-time upper bound; external QEMU+fio reference runs remain the real NAND-path scaling validator.
- **REQ-123 — CPU + DRAM budgets.** `bench_run` brackets the workload with a local `system_monitor` (getrusage-backed); `cpu_util_pct` is a real reading, not a hardcoded 0. The run_all gate splits REQ-123 into two rows: `REQ-123` (CPU ≤ 50% under a bursty-dispatch probe) and `REQ-123-DRAM` (process RSS ≤ 1024 MB under the same probe).
- **`hfsss_nbd_server` sampler wiring.** New process-wide `system_monitor` constructed at startup (1s poll, default POSIX CPU+RSS callbacks). Thread-count gauge updated via `__atomic_store_n` when the MT FTL pool spins up. `atexit()` teardown covers all early-return paths.

## Coverage delta

Grand Total: **146 ✅ / 14 ⚠️ (82.0%)** — up from 144/18 (80.9%).

- Performance Requirements: 5/3 → **7/1** (62.5% → 87.5%).
- Core Subtotal: 105/20 → **107/18** (76.1% → 77.5%).

## Test plan

- [x] `test_perf_validation`: 88 → **100** (+12 assertions: direct `bench_run.cpu_util_pct` probe + REQ-121/122/123/123-DRAM pass + target pins + 12-row inventory).
- [x] `make build/bin/hfsss-nbd-server` compiles cleanly.
- [x] Full `make test`: **0 FAIL**.

## Design notes

- REQ-122 is now actually measured — the gate can fail on a real worker-path regression, not only when someone edits the Amdahl constant.
- REQ-123's CPU probe uses a burst+idle window because a synthetic hot loop saturates ~100% CPU and misses the PRD intent ("controller firmware under bursty load"). The probe shape is documented in-code.
- The NBD server's thread-count gauge uses relaxed atomics — it's a reporting counter, not a synchronization variable.

### DIFF
--- a/docs/REQUIREMENT_COVERAGE.md
+++ b/docs/REQUIREMENT_COVERAGE.md
@@ -33,11 +33,11 @@ This document analyzes the coverage of the 178 requirements from the Requirement
 | HAL | 12 | 12 | 0 | 0 | 0 | 100% | ↑ REQ-063 AER + REQ-064 PCIe link state + REQ-069 byte-level config space (LLD_13) |
 | Common Services | 24 | 20 | 2 | 2 | 0 | 83.3% | ↑ REQ-085 SPSC ring + REQ-087 system resource monitor on top of boot/power/OOB/SMART/trace |
 | Algorithm Task Layer (FTL) | 22 | 19 | 1 | 2 | 0 | 86.4% | ↑ +6 (cmd state machine, retries, flow ctl, wear monitor, Error Log Page) |
-| Performance Requirements | 8 | 5 | 0 | 3 | 0 | 62.5% (62.5% partial) | ↑ REQ-116..120 perf targets asserted in run_all report + gated by regression suite |
+| Performance Requirements | 8 | 7 | 0 | 1 | 0 | 87.5% (87.5% partial) | ↑ REQ-116..120 + REQ-122 Amdahl scalability + REQ-123 bursty-load CPU probe gated by regression suite |
 | Product Interfaces | 8 | 4 | 3 | 1 | 0 | 50.0% | ↑ +4 (/proc, hfsss-ctrl, YAML, persistence) |
 | Fault Injection | 3 | 2 | 1 | 0 | 0 | 66.7% (100% partial) | ↑ registry + power hook + NAND read/program/erase fault gates landed |
 | System Reliability | 4 | 3 | 0 | 1 | 0 | 75.0% | ↑ REQ-088 P99.9 latency anomaly detector |
-| **Core Subtotal** | **138** | **105** | **13** | **20** | **0** | **76.1%** (85.5% partial) | ↑ from 50.0% |
+| **Core Subtotal** | **138** | **107** | **13** | **18** | **0** | **77.5%** (87.0% partial) | ↑ from 50.0% |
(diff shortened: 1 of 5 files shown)


AlvinYangZF:
All 7 Codex findings verified and addressed in commit \`17aaf30\`.

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | High | REQ-123 CPU-only probe left DRAM unchecked despite PRD saying "CPU/DRAM" | Split into \`REQ-123\` (CPU ≤ 50%) + \`REQ-123-DRAM\` (RSS ≤ 1024 MB). Both sampled by the same \`system_monitor\` during the probe window. Report contract now pins a 12-row inventory |
| 2 | High | REQ-122 Amdahl check was a compile-time constant — unable to regress | Replaced with **measured** ratio \`iops(8) / (8 * iops(1))\` from back-to-back \`bench_run\` invocations. Amdahl remains as design-time upper bound in the code comment |
| 3 | Medium | `bench_run` file-scope `g_bench_thread_count` raced concurrent callers | Pass count via `cb_ctx` pointer; static removed |
| 4 | Medium | bench_run monitor leaked on early-return paths | Every failure leg (threads alloc / started alloc / pthread_create fail / pthread_join fail) now calls `system_monitor_cleanup` before bailing |
| 5 | Medium | NBD server: `init` succeeded + `start` failed left a dangling mutex; `atexit` return value ignored; thread-count gauge never updated | `cleanup` on start-fail; atexit return checked with synchronous fallback; `__atomic_store_n` updates the gauge when MT FTL pool comes up; callback uses matching `__atomic_load_n` |
| 6 | Low | Docs self-contradicted: `TRACEABILITY_MATRIX.md` detail still said REQ-122/123 Not Implemented while summary said Implemented; stale "Remaining Major Gaps" + "Near-term priorities" lines in `REQUIREMENT_COVERAGE.md` | Reconciled all three doc locations. Near-term priorities retired the stale perf items and now points at the next core gaps |
| 7 | Low | No direct test for `bench_run.cpu_util_pct`; no pass/target pins for REQ-123-DRAM | `test_bench_run_populates_cpu_util` asserts > 0 and sane upper bound. REQ-123-DRAM row gets pass + target pins in `test_validation_run_all` |

### Verification

- `test_perf_validation`: 88 → **100** (+12 assertions).
- Full `make test`: **0 FAIL**.

Codex also noted commit/PR progress-label wording ("Tier-F left outstanding" / "Tier-H") in the earlier PR body. The PR body has been retitled via `gh pr edit`; commit `c5768e2`'s body is merged-history-adjacent, left as-is to avoid destructive force-push.

PR #99, opened 2026-04-21
feat(media,controller): multi-plane concurrency + per-channel worker runtime

AlvinYangZF:
## Summary

Adds verified multi-plane concurrency evidence (REQ-042) and an additive per-channel worker runtime (REQ-044) with async-completion scaffolding (REQ-045 Partial). Lifts grand-total coverage **82.0 % → 83.1 %** (146 → 148 ✅, 14 → 13 ⚠️, 18 → 17 ❌).

### REQ-042 — NAND multi-plane concurrency (⚠️ → ✅)
`cmd_engine` already iterates `plane_mask` and advances per-plane EAT while draining array wall-clock exactly once. Missing piece was verification. `tests/test_media_multi_plane_concurrency.c` adds four coverage blocks:
- Multi-plane PROG wall-clock ≤ 1.5× single-plane baseline; sequential two-plane PROG ≥ 1.8× baseline; mp2 strictly faster than 2× sequential. SLC-timing run: single=203 µs, 2-plane=202 µs, 4-plane=215 µs, seq×2=403 µs.
- Per-plane EAT independence: targeted planes share one post-op EAT; untargeted planes stay unmoved.
- Profile plane-cap enforcement: ONFI (cap 4) accepts 4-plane PROG; Toggle (cap 2) rejects 3-plane with `HFSSS_ERR_INVAL`.
- Multi-plane READ: pre-read EATs aligned via a single multi-plane PROG so the new `plane0_delta == plane1_delta` assertion genuinely proves parallel per-plane READ charge; payload fidelity verified.

### REQ-044 — per-channel command queue + thread (⚠️ → ✅)
New `channel_worker` runtime (`include/controller/channel_worker.h`, `src/controller/channel_worker.c`): one `pthread` per logical channel, an SPSC request ring (reusing REQ-085 infrastructure), non-blocking `channel_worker_submit` returning `HFSSS_ERR_BUSY` when the ring is full, channel-id enforcement so a worker only drives its own channel's media ops, and a stop lifecycle that rejects late submits and drains in-flight work before the thread exits.

### REQ-045 — async completion scaffolding (❌ → ⚠️ Partial)
`channel_cmd` carries an `on_complete` callback and an atomic `done` flag; `channel_cmd_wait` provides a poll-style barrier for submitter threads. The worker publishes `done` with release ordering BEFORE firing the callback so a reclaiming callback does not race the worker's trailing store. Callers pick either callback or wait as the reclamation hook, not both.

**Scope note**: the PRD wording for REQ-045 calls for a lock-free completion queue with per-command timestamp, actual latency, and batch-drain behaviour. This PR delivers the callback/done scaffolding only. Full CQ remains a follow-up — tracked in REQUIREMENT_COVERAGE.md near-term priorities.

### Design choice — additive runtime, not a hot-path retarget
`channel_worker` is an additive layer. The existing synchronous `media_nand_*` call path used by FTL / GC / controller flows is untouched; callers opt into async submission by constructing a `channel_cmd`. Retargeting FTL on top of `channel_worker` is intentionally deferred — realising the async benefit in the hot path is a bigger change that deserves its own branch and benchmarks.

## Review resolution

Independent review surfaced seven findings; all addressed in the follow-up commit. Highlights:
- done ordering flipped to publish-before-callback so reclaiming callbacks are safe
- submit-after-stop now returns BUSY; worker drains before exit
- per-worker channel-id enforced with a new cross-channel rejection test
- ring-full test rewritten to flood a LIVE worker (no internal-state manipulation)
- REQ-045 demoted to Partial to reflect the CQ gap honestly
- multi-plane READ test pre-aligns plane EATs so the new delta-equality assertion genuinely proves parallel charge
- counter documentation rewritten to describe lifetime rather than instantaneous depth

## Test plan

- [x] `test_media_multi_plane_concurrency` — 48/48 pass (parallel timing + EAT independence + plane cap + aligned READ parity).
- [x] `test_channel_worker` — 42/42 pass (submit+cb, erase/program/read roundtrip, ring-full under real load, submit-after-stop, cross-channel rejection, two-channel isolation, NULL hardening).
- [x] Full `make test` regression — 0 real failures across the existing suite.

## Coverage delta

| Module | Before | After |
|---|---|---|
| Media Threads | 75.0 % (15/20) | 85.0 % (17/20) |
| Core Subtotal | 77.5 % | 79.0 % |
| **Grand Total** | **82.0 %** | **83.1 %** |


AlvinYangZF:
Addressed the seven findings from the earlier independent review in commit `30d0532`. Summary:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | HIGH | `on_complete` fired before `done` store → UAF risk if callback reclaims | Flipped order: worker now stores `done` with release BEFORE invoking callback. Header contract rewritten to state callback and wait are alternative reclamation paths, not both. |
| 2 | HIGH | Submit-after-stop could orphan commands; ring-full test manipulated internals | `channel_worker_submit` now acquires the stop flag up front and rejects with `HFSSS_ERR_BUSY`; worker loop drains any racily-submitted commands before breaking. Added `test_submit_after_stop_returns_busy`; rewrote `test_full_ring_returns_busy` to flood a live worker with a 64-cmd burst into a cap-2 ring and drain via the completed counter. |
| 3 | LOW | `submitted - completed` advertised as in-flight depth but can underflow | Struct comment rewritten — counters are lifetime-only, use them for rate monitoring, not saturation. |
| 4 | HIGH | Worker did not enforce own channel_id; could drive any channel's ops | `dispatch_one` now returns `HFSSS_ERR_INVAL` when `cmd->ch != w->channel_id`. Added `test_cross_channel_cmd_rejected` proving the cmd still completes with INVAL status. |
| 5 | HIGH | REQ-045 closure overstated — PRD wants lock-free CQ with timestamp / latency / batch | Flipped REQ-045 back to ⚠️ Partial in both matrices; note cites the callback/done scaffolding that landed and names the CQ follow-up. Coverage counters recomputed (Grand 83.1 %, was 83.7 %). |
| 6 | MEDIUM | READ delta assertion passed on equal finals despite unequal deltas (1.7 ms vs 31 µs) | Pre-read EATs now aligned via a single multi-plane PROG. Added `plane0_delta == plane1_delta` assertion that genuinely proves per-plane parallel charge. |
| 7 | NIT | Commit body used the banned "Tier" wording | Fresh fix commit + PR body avoid the wording. Earlier commit body left as-is to avoid a history rewrite. |

Test evidence:
- `test_media_multi_plane_concurrency` 48/48 (+1 from the new delta-equality assertion)
- `test_channel_worker` 42/42 (+7 from stop-rejection + cross-channel + ring-full-under-load split + cleanup bookkeeping)
- Full `make test` regression clean (0 real failures)

Coverage after fixes:
- Media Threads 85.0 % (17 ✅ / 3 ⚠️ / 0 ❌)
- Core Subtotal 79.0 % (109 ✅ / 12 ⚠️ / 17 ❌)
- Grand Total **83.1 %** (148 ✅ / 13 ⚠️ / 17 ❌)

PR #100, opened 2026-04-22
fix(test): align sanitize expectations with NVMe SANACT semantics

AlvinYangZF:
## Summary

Clears the 3 pre-existing sanitize test failures surfaced by PR #99's post-merge E2E sweep (2 in DI-007, 1 in NC-013). Root cause was **test-expectation bugs**, not an implementation gap — `nvme_uspace_sanitize` already dispatches the four NVMe §5.22 SANACT codes per spec.

## What was wrong

Both DI-007 and NC-013 asserted that every SANACT leaves user data unreadable (NOENT). NVMe §5.22 actually says:

| SANACT | Action | Post-state |
|---|---|---|
| 0x01 | EXIT_FAILURE | Not a sanitize op; data intact |
| 0x02 | BLOCK_ERASE | Data erased (simulator drops mapping → NOENT) |
| 0x03 | OVERWRITE | Data rewritten with pattern (default zeros) → OK + zero bytes, not NOENT |
| 0x04 | CRYPTO_ERASE | Keys destroyed (simulator drops mapping → NOENT) |

DI-007 looped `sanact=1..3` expecting NOENT for all three — wrong for 0x01 and 0x03. NC-013 made the same mistake on 0x01.

## What this PR does

- **`systest_data_integrity::DI-007`** — refactored the hand-rolled loop into a case table driven by `enum sanact_expect {PRESERVE, NOENT, ZERO_FILL}`. Each sanact gets its own write-sanitize-verify cycle with an action-specific post-read assertion. Added sanact=4 CRYPTO_ERASE coverage.
- **`systest_nvme_compliance::NC-013`** — rewritten as a dedicated write+sanitize+read block per SANACT code with pattern-fidelity assertions for PRESERVE and ZERO_FILL modes. Added sanact=4.
- **`docs/REQUIREMENT_COVERAGE.md`** — V3.0 note updated: REQ-163 detail row already correctly showed ✅, the "demoted to partial" sentence in the summary was a note-level stale claim. Replaced with a pointer to the real implementation + passing tests.

## Test evidence

- `systest_nvme_compliance`: **70 → 75 assertions**, 0 failed (was 69/70 with 1 fail)
- `systest_data_integrity`: **50 → 53 assertions**, 0 failed (was 48/50 with 2 fails)
- Full `make test` regression clean aside from a pre-existing REQ-122 live-scalability flake (measured 69–72% against a 70% threshold; unrelated to this PR).

## No source changes

`src/pcie/nvme_uspace.c::nvme_uspace_sanitize` is untouched — its switch already implements all four SANACT codes correctly.


AlvinYangZF:
Addressed the Codex review findings in commit \`d212d2b\`.

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | HIGH | DI-007 / NC-013 / PR copy framed post-SANACT expectations as "per NVMe §5.22 spec" when the spec actually leaves deallocated / overwritten / crypto-erased content loose | Docstrings and V3.0 note rewritten to describe each expectation as the simulator's concrete observable that satisfies the spec's underlying guarantee (old data unrecoverable / host-pattern written / no-op). Assertions unchanged — they pin the simulator's modeled outcomes. |
| 10 | LOW | NC-013 seed writes for sanact=2/3/4 were not return-code-checked, so a write-path regression could vacuously pass | Every sanact block now asserts the seed write's return code before issuing sanitize. NC-013: 75 → 78 assertions. |
| 7 | NIT | \`overwrite_fail\` in DI-002 silenced via \`__attribute__((unused))\` instead of removed | Dead counter and its lone \`else\` arm deleted. DI-007 assertion count unchanged at 53. |
| 5 | MEDIUM | LLD_19 line 42 labels REQ-163 as "key storage" while TRACEABILITY_MATRIX labels it "secure erase" | Out of scope for this PR — LLD-internal mislabel, not created by this change. Flagging as a tracked doc-hygiene follow-up. |
| 2, 3, 4, 6, 8, 9 | NIT | Codex "no issue found" | — |

Test evidence:
- systest_nvme_compliance: 75 → **78** assertions (+3 from new seed-write rc asserts), 0 FAIL
- systest_data_integrity: unchanged at **53/53** (dead counter removal is assertion-neutral)
- Full \`make test\` regression: **3668 PASS / 0 FAIL**

PR #102, opened 2026-04-22
fix(ci): link -lm into new systest binaries for Linux builds

AlvinYangZF:
## Summary

Linux CI broke after PR #101 merged. Root cause: the two new systest rules (`systest_nvme_cli_compat`, `systest_fio_compat`) link `libhfsss-common.a` (which uses `exp()` / `log()` from `uplp.c`) without `-lm`. macOS links libm by default so the local build stayed green; ubuntu-latest failed with `undefined reference to 'exp' / 'log'`.

All other rules linking `libhfsss-common.a` already include `-lm`. Aligned the two new rules with the rest of the Makefile.

## Why this slipped past local testing

macOS `ld` resolves libm symbols implicitly against the system library; Linux `ld` does not. The local `make test` on macOS passed 3697 / 0 even though the Linux toolchain would have failed at link time. This is a known macOS / Linux divergence, not an issue with the systest logic itself.

## Test plan

- [x] Local rebuild of both affected binaries on macOS passes (`systest_nvme_cli_compat`: 21 / 21, `systest_fio_compat`: 8 / 8)
- [x] Diff is a two-line `-lm` insertion, consistent with every other rule that links `libhfsss-common.a`
- [ ] Await ubuntu-latest + macos-latest CI on this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@ CFLAGS += -Iinclude -Iinclude/common -Iinclude/media -Iinclude/hal -Iinclude/ftl
 CFLAGS += -MMD -MP
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
-    LDFLAGS = -lpthread -lrt
+    LDFLAGS = -lpthread -lrt -lm
 else
-    LDFLAGS = -lpthread
+    LDFLAGS = -lpthread -lm
 endif
 
 # Coverage build variant (GCC --coverage instrumentation)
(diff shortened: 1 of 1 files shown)


AlvinYangZF:
Expanded PR scope with commit 9ff66cb to address both LOW follow-ups from review:

**Finding 1 (other `-lhfsss-common` rules missing `-lm`)**: Rather than patch seven individual rules, centralized `-lm` into `LDFLAGS` for both the Linux and non-Linux branches of the OS conditional. Every rule that links `-lhfsss-common` now picks up libm automatically; the pattern cannot recur for future binaries. macOS `ld` tolerates the extra `-lm` (resolves to libSystem / system libm) with no effect.

**Finding 2 (`all:` references `$(FTL_MFC_REPRO)` before assignment)**: Moved the `FTL_MFC_REPRO = $(BIN_DIR)/ftl_mfc_repro` assignment from line 619 up to alongside the other `$(BIN_DIR)/*` variable declarations at line ~172. GNU make expands prerequisite lists when the rule is parsed, so the forward reference was silently expanding to empty. Verified:

- Before: `make -pn | grep ^all:` produced no `build/bin/ftl_mfc_repro` entry.
- After: `make -pn | grep ^all:` includes it; `make -j4 all` builds it as expected.

`make -j4 all` rebuilds cleanly on macOS. Every previously `-lm`-less rule now links.

PR #111, opened 2026-04-25
feat(channel_worker): per-cmd timestamps + opt-in completion queue (REQ-045)

AlvinYangZF:
## Summary
Close REQ-045 (Media Threads) by extending the existing `channel_worker` runtime with:
- per-command wall-clock timestamps (`submit_ts_ns` / `complete_ts_ns`) recorded unconditionally
- an opt-in lock-free SPSC completion queue selectable at `channel_worker_init` (`cq_capacity` arg, `0` preserves current behavior)
- non-blocking batch drain via `channel_worker_drain`
- three-path mutual exclusion (legacy `channel_cmd_wait` poll, `on_complete` callback, CQ batch drain) enforced at submit / wait time

CQ-full applies back-pressure (worker spin-retries with `sched_yield`) so no completion is ever dropped — back-pressure flows from CQ into the submit ring instead. Scope follows the REQ-121 / REQ-048 / REQ-137 convention: mechanism + harness shipped; tier-2 controlled CQ benchmark and tier-3 retarget of the FTL/HAL hot path through `channel_worker` are explicit follow-ups (see spec Section 9).

## Changes
- `include/controller/channel_worker.h` — `channel_cmd` gets `submit_ts_ns` / `complete_ts_ns` / `owner`; `channel_worker` gets `cq_enabled` / `cq`; `channel_worker_init` takes a new `cq_capacity` arg; `channel_worker_drain` declared; per-function docblocks aligned with the new three-path contract.
- `src/controller/channel_worker.c` — submit-side timestamp + owner + on_complete-vs-CQ rejection; `channel_cmd_wait` rejects in CQ mode via owner back-pointer; worker stamps `complete_ts_ns` and spin-retries CQ push before publishing `done`; live `channel_worker_drain`.
- `tests/test_channel_worker.c` — 8 existing call sites pass `0` for new arg; 8 new test cases cover all three paths.
- `docs/REQUIREMENT_COVERAGE.md` — REQ-045 ⚠️→✅; subtotal updates.
- `docs/superpowers/specs/2026-04-24-req-045-completion-queue-design.md` (new) — full design spec including tier-2/3 follow-up roadmap.
- `docs/superpowers/plans/2026-04-24-req-045-completion-queue.md` (new) — five-task TDD implementation plan.

## Scope boundary
- Mechanism + tests: shipped.
- Tier-2 controlled CQ benchmark harness: follow-up (spec Section 9.1).
- Tier-3 retarget FTL/HAL hot path: follow-up (spec Section 9.2).
- Blocking drain / modeled-EAT latency / per-channel CQ stats: explicitly deferred (spec Section 3.2).

## Test plan
- [x] `make test` — full regression 0 [FAIL]; 8 new channel_worker test cases pass
- [x] `cq_capacity=0` path is binary-compatible with current behavior; existing tests unchanged in semantics
- [x] CQ basic drain pops in submit order; timestamps populated on every cmd
- [x] CQ-full produces back-pressure into submit ring (`HFSSS_ERR_BUSY`); no dropped completions
- [x] Three-path enforcement: rejecting on_complete at submit, rejecting wait at entry, rejecting drain on cq=0 worker
- [ ] Codex review pass (queued for user to trigger)


src/controller/channel_worker.c
```diff
@@ -31,12 +31,43 @@ static int dispatch_one(struct channel_worker *w, struct channel_cmd *cmd)
...
+     * been signaled. On stop we bail out without setting done; the
+     * cmd stays in the request-ring drain semantics handled by the
+     * worker loop and cleanup path (i.e. it's discarded along with
+     * any other in-flight work, same as the legacy path).
+     *
+     * on_complete is rejected at submit-time when CQ is enabled, so
+     * the legacy callback branch below is unreachable in CQ mode. */
+    if (w->cq_enabled) {
+        for (;;) {
+            int rc = spsc_ring_tryput(&w->cq, &cmd);
```
AlvinYangZF, line 66: This publishes `cmd` into the completion queue before the worker stores `cmd->done = 1` below. Because `spsc_ring_tryput` releases the CQ slot immediately, a fast CQ consumer can drain this pointer, observe `done == 0`, or legitimately reclaim/reuse the command while the worker is still about to write `done` on line 71. That violates the documented “completed cmd” / ownership-transfer contract and can turn into a data race or use-after-free in real CQ consumers.

Please publish all per-command completion state before enqueuing to the CQ. In practice, CQ mode should stamp `complete_ts_ns`, set `status`, store `done = 1` with release semantics, and only then release the pointer through `spsc_ring_tryput`; after the CQ enqueue, the worker should not write to `cmd` again. It would also be worth adding a test that drains and immediately asserts `done == 1` / reuses the command to cover this interleaving.

AlvinYangZF (commented):
I found one correctness issue that should be fixed before merging. The CQ path publishes the command pointer to the consumer before the worker has finished writing the command's completion state, so the completion queue can hand ownership to the caller too early.

src/controller/channel_worker.c
AlvinYangZF: Confirmed fixed in `48c7638`. CQ mode now publishes `done = 1` and increments the completion counter before releasing the command pointer through `spsc_ring_tryput`, and the worker returns immediately after a successful CQ enqueue so it no longer touches `cmd` after ownership is transferred. The new `test_cq_drain_then_reuse_race` also covers the drain-immediate-read/reuse case I was concerned about.

I re-ran `make build/bin/test_channel_worker` and `./build/bin/test_channel_worker`; the targeted suite passes.

PR #112, opened 2026-04-26
feat(channel_worker): tier-2 CQ latency bench harness (REQ-045)

AlvinYangZF:
## Summary

Tier-2 follow-up to PR #111 (REQ-045 tier-1 completion queue). Adds a controlled benchmark harness that exercises the opt-in CQ under a mixed read/program workload and reports submit-to-complete latency percentiles. **Opt-in only — not wired into `make test`.** No production-path retarget (that's tier-3).

Spec reference: `docs/superpowers/specs/2026-04-24-req-045-completion-queue-design.md` §9.1.

## Changes

- **New** `tests/bench_channel_worker_cq.c` (~750 LOC)
  - One `channel_worker` per channel, `cq_capacity=256`, one producer + one consumer thread per channel
  - Producer issues 70/30 R/W mix against pre-erased blocks; recycles blocks via ERASE submissions when a block fills up
  - Consumer drains the CQ in batches of 64, records `complete_ts_ns - submit_ts_ns` per cmd
  - Aggregates across channels, prints count, min/max, mean, p50, p99, p99.9 + wall time + ops/sec
  - When `STRESS_RESULTS_FILE` is set, also writes a `key=value` summary (same shape as `tests/stress_stability.c`)
- **Modified** `Makefile`
  - New `BENCH_CQ` binary, build recipe matching the other channel-worker test
  - New `bench-cq` phony target, env-var passthrough (`STRESS_RESULTS_FILE`, `BENCH_OPS`, `BENCH_RATE_OPS_PER_SEC`)
  - Listed under `make help` as ` Run REQ-045 tier-2 CQ latency bench harness (opt-in)`
  - **Not** added to `make test`

## Workload knobs

| Env var | Default | Purpose |
|---|---|---|
| `BENCH_OPS` | 10000 | ops per channel |
| `BENCH_RATE_OPS_PER_SEC` | 0 (unpaced) | producer pacing rate |
| `STRESS_RESULTS_FILE` | unset | path for key=value scrape file |

## PASS criterion

Every submitted op observed by consumer **AND** aggregate p99 < **100 ms**.

The 100 ms budget is intentionally a 3× headroom over the natural unpaced p99 (~30 ms) seen on this in-memory simulator. The producer saturates the 256-deep SPSC ring on first turn so each cmd accrues queue-wait latency — that's expected backpressure, not a defect. The budget is sized to catch order-of-magnitude regressions (deadlock, unbounded spin, livelock) without flaking on natural queueing. Tighten by passing `BENCH_RATE_OPS_PER_SEC` to pace the producer.

## Representative output

\`\`\`
total_ops        : 20000
wall_time_sec    : 0.986
ops_per_sec      : 20277.0
latency_min_ns   : 202000
latency_p50_ns   : 25145000
latency_p99_ns   : 29496000
latency_p999_ns  : 30425000
latency_max_ns   : 31139000
latency_mean_ns  : 24946545.8
result           : PASS
\`\`\`

## Scope boundary (explicit)

In scope:
- Standalone bench harness on \`channel_worker\` API only
- p50 / p99 / p99.9 latency reporting
- key=value scrape summary

Out of scope (tier-3, separate plan):
- FTL / HAL hot-path retarget onto \`channel_worker\`
- fio-visible latency export
- Production NBD / NVMe path changes

## Test plan

- [x] \`make -j12 all\` — clean build, \`build/bin/bench_channel_worker_cq\` produced
- [x] \`make bench-cq\` — runs to completion, prints latency report, exit 0 on PASS
- [x] \`STRESS_RESULTS_FILE=/tmp/bench_cq_results.txt make bench-cq\` — key=value file written with documented keys
- [x] \`make test\` — regression unchanged (tripped the known REQ-122 wall-clock flake on first run; clean on re-run, confirming pre-existing flake unrelated to this change)
- [ ] CI build matrix (ubuntu-latest, macos-latest, kernel-module) green
- [ ] Sanity-run \`make bench-cq\` once locally on Linux to confirm portability of \`clock_gettime(CLOCK_MONOTONIC)\` + \`pthread\` paths

### DIFF
--- a/tests/bench_channel_worker_cq.c
+++ b/tests/bench_channel_worker_cq.c
@@ -20,16 +20,27 @@
  *   BENCH_RATE_OPS_PER_SEC  — pacing rate (0 = unpaced; default 0).
  *   STRESS_RESULTS_FILE     — when set, key=value summary is appended.
  *
- * PASS criterion: every submitted op was observed by the consumer AND
- * the aggregate p99 latency is below the documented budget. Latency
- * here is wall-clock submit→complete, so an unpaced producer that
- * saturates the SPSC ring naturally accrues queue-wait at the head of
- * each op; with ring depth of 256 and the modeled NAND op time the
- * harness observes p99 in the low tens of milliseconds. The budget is
- * deliberately set to 100 ms so the harness catches order-of-magnitude
- * regressions (e.g. an unbounded spin or a latent deadlock surfacing
- * as a multi-second tail) without flaking on natural queue-depth
- * backpressure. Exit 0 on PASS, non-zero otherwise.
+ * PASS criterion (all four must hold):
+ *   - every submitted op was observed by the consumer through the CQ;
+ *   - every observed op completed with status == HFSSS_OK (failed_ops == 0);
+ *   - no producer/consumer hit a fatal harness error;
+ *   - aggregate p99 latency over the OK-only sample set is below the
+ *     documented budget.
+ *
+ * Failure status from the worker is the gate that distinguishes a
+ * working CQ from a CQ that merely delivers commands. Failed ops are
+ * counted and reported but excluded from the latency percentile pool,
+ * because failures often complete quickly and would skew the
+ * distribution downward and hide real regressions.
+ *
+ * Latency here is wall-clock submit→complete, so an unpaced producer
+ * that saturates the SPSC ring naturally accrues queue-wait at the
+ * head of each op; with ring depth of 256 and the modeled NAND op
+ * time the harness observes p99 in the low tens of milliseconds. The
+ * budget is deliberately set to 100 ms so the harness catches
+ * order-of-magnitude regressions (e.g. an unbounded spin or a latent
+ * deadlock surfacing as a multi-second tail) without flaking on
+ * natural queue-depth backpressure. Exit 0 on PASS, non-zero otherwise.
  *
  * Standalone — not wired into `make test`. Build via `make bench-cq`.
  */
@@ -105,8 +116,11 @@ struct bench_channel {
     _Atomic int            prod_done;
     _Atomic u64            submitted;
     _Atomic u64            completed;
+    _Atomic u64            failed_ops;       /* drained cmds whose status != HFSSS_OK */
+    _Atomic int            first_fail_status;/* first observed non-OK status, or 0 */
     _Atomic int            fatal;            /* 1 if producer/consumer hit a fatal error */
-    /* Latency samples — owned by the consumer, sized to target_ops. */
+    /* Latency samples — owned by the consumer. Only successful (HFSSS_OK)
+     * completions are recorded; sized to target_ops as the worst-case bound. */
     u64                   *samples;
     u32                    sample_count;
 };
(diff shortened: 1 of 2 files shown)


tests/bench_channel_worker_cq.c
```diff
@@ -0,0 +1,749 @@
...
+                atomic_load(&bc->submitted) == popped_total) {
+                break;
+            }
+            sched_yield();
+            continue;
+        }
+        for (int i = 0; i < n; i++) {
+            struct channel_cmd *cc = batch[i];
+            struct bench_cmd   *cmd = (struct bench_cmd *)cc;
+            u64 lat = cc->complete_ts_ns - cc->submit_ts_ns;
```
AlvinYangZF: The consumer records latency and counts every drained command as observed, but it never checks `cc->status`. That lets the benchmark report `PASS` even if the worker completed commands with `HFSSS_ERR_INVAL`, `HFSSS_ERR_IO`, or another media failure instead of actually executing the intended read/program/erase workload. In that case `total_ops == expected_total` and the p99 budget can still pass, especially because failed commands may complete quickly.

Please treat any non-`HFSSS_OK` command status as a benchmark failure, or at least track/report failed-op counts and include `failed_ops == 0` in the PASS criterion. That would make the harness validate successful CQ completions rather than only CQ delivery.

AlvinYangZF (commented):
I found one benchmark-correctness issue that should be addressed before relying on this harness as the REQ-045 tier-2 signal.

AlvinYangZF:
Addressed in e1264ed.

The consumer drain loop now checks `cc->status`:
- `HFSSS_OK` samples go into the latency percentile pool (per-channel `sample_count`)
- non-OK results increment a new `failed_ops` counter and capture the first failed status code into `first_fail_status` for triage
- both still increment the drained counter and free the heap-allocated cmd, so producer-done loop termination is unchanged

PASS criterion is now `!any_fatal && failed_ops == 0 && observed_ops == expected_total && ok_ops == expected_total && p99 < budget`. Failed samples are excluded from the percentile pool so a partial-failure run cannot mask the latency tail by skewing p99 downward.

The reporter and key=value summary surface `observed_ops`, `ok_ops`, `failed_ops`, `first_fail_status`. `total_ops` was renamed to `ok_ops` (semantics-preserving on a healthy run). Header docstring rewritten to spell out the four PASS conditions explicitly.

tests/bench_channel_worker_cq.c
```diff
@@ -0,0 +1,803 @@
...
+
+        wait_until(deadline_for(start, bc->rate_ops_per_sec, i));
+
+        /* Submit with retry on BUSY (CQ back-pressure flows through to
+         * the submit ring). yield-spin keeps the producer responsive
+         * to cleanup. */
+        for (;;) {
+            int rc = channel_worker_submit(&bc->worker, &cmd->ccmd);
+            if (rc == HFSSS_OK) break;
+            if (rc == HFSSS_ERR_BUSY) {
```
AlvinYangZF, line 385: The BUSY retry loop should also observe the harness shutdown/fatal signal. If teardown calls `channel_worker_stop()`, `channel_worker_submit()` can keep returning `HFSSS_ERR_BUSY`; this loop currently just yields and retries forever, so `teardown_channels()` can hang while joining the producer thread. Please check `bc->fatal` (or otherwise detect stop) in this branch, free the current `cmd`, set `prod_done`, and exit the producer cleanly.

AlvinYangZF (commented):
I found one additional harness-stability issue in the latest revision.

AlvinYangZF:
Addressed in d6d70ab.

The producer's BUSY-yield loop now checks `bc->fatal` before each yield. If teardown has signalled stop, the producer frees the unsubmitted cmd and `goto out` which sets `prod_done` and returns — `pthread_join` in `teardown_channels()` no longer hangs.

Outer per-iteration fatal check is preserved (covers the case where stop fires between iterations). Consumer side reviewed and unaffected: `channel_worker_drain` is non-blocking and the consumer loop already polls `bc->fatal` at the top of every iteration.

Healthy-path bench-cq still PASS at 20k ops; latency distribution unchanged (p50 25.2ms / p99 29.5ms / p99.9 30.4ms).

PR #115, opened 2026-04-26
fix(ftl): async-FTL race via retry-with-backoff hardening (FTL only)

AlvinYangZF:
## Status: ready for review — pre-checkin 8/8 PASS, full unit suite passes

Investigation + fix for the async-pipeline FTL second race that failed
`make pre-checkin` on fio 010-014. Original "TAA L2P race in gc.c"
hypothesis was empirically refuted via Path A reset and replaced with
the correct one: cmd_engine fail-fast + FTL 3-retry-without-backoff
caused thread-level starvation under heavy contention.

A cv-based wait-on-busy in cmd_engine was attempted and reverted (see
"Iterations" below). The landed fix is FTL-only.

## TL;DR

Pre-checkin failed with NVMe SCT=0x2 SC=0x80 Write Fault. Root cause:

1. `cmd_engine.c::engine_submit` returns `HFSSS_ERR_BUSY` immediately
   when target die is in `DIE_*_ARRAY_BUSY`. **This is the documented
   hardware contract that IS-04 / IS-06 / SUSPEND tests assert; we
   keep it.**
2. FTL retried 3× back-to-back with no backoff. tProg is 900-1300 µs
   and 3 retries cleared in microseconds, so all 3 hit the same busy
   state.
3. After 3 fails, `ftl_write_page_mt` marked the block CLOSED and
   reset `cwb->block`. Next host write to the same CWB allocated a
   fresh block on the same hot die — burning blocks per attempt.
4. NBD CQ saw `slot->status == HFSSS_ERR_BUSY` → emitted EIO. QEMU's
   `nvme_aio_err` mapped EIO-on-write to SCT=0x2 SC=0x80 Write Fault.

Fix: FTL retry-budget hardening, all in `src/ftl/ftl.c`:

- **Backoff via `nanosleep(100 µs)` on transient BUSY/AGAIN.**
  `sched_yield()` is too short — TLC tProg is 900-1300 µs.
- **Retry budget 3 → 2048.** `2048 × 100 µs ≈ 205 ms` covers
  worst-case starvation (fio-012 bs=128k iodepth=16 generates ~64
  sub-writes per die at peak). Well within NVMe 30 s timeout.
- **Don't mark blocks CLOSED on transient BUSY/AGAIN.** Page never
  programmed → block is fine; preserve it across retries instead of
  burning fresh blocks per failure.

## Iterations on this branch

| Approach | Result | Outcome |
|---|---|---|
| 64 retries × 50 µs | fio-010/011 PASS, 012/013/014 FAIL | budget too small |
| 512 retries × 100 µs | fio-010/011/014 PASS, 012 FAIL | borderline; trace overhead masks the issue |
| **cmd_engine cv wait-on-busy** | **fio-012 PASS but breaks IS-04 / IS-06 / test_media SUSPEND tests** | **reverted (commit `32e4611`)**; cv requires a separate blocking-vs-nonblocking API split, out of scope |
| **2048 retries × 100 µs** ✅ | **pre-checkin 8/8 PASS, full unit suite passes, reproducer err_rate=0** | **landed (commit `b8b6dae`)** |

## Validation

| Metric | Before | After |
|---|---|---|
| `ftl_mfc_repro --threads 16` read err_rate | 1.07e-01 | **0** |
| `ftl_mfc_repro --threads 128` read err_rate | 6.78e-02 | **0** |
| fio-012 single case (no trace) | FAIL (SCT=0x2 SC=0x80) | PASS |
| `make pre-checkin` (8 cases + spdk skip) | 010-014 FAIL | **8/8 PASS** |
| `make test` IS-04 / IS-06 / SUSPEND assertions | pass | pass (cv approach broke these — reverted) |
| `make test` overall | pass | pass (1 known-flaky test_media_multi_plane_concurrency, pre-existing, pre-existing-flaky on second run) |

## Files touched (production code)

- `include/common/io_err_trace.h` — new env-gated trace helper for
  diagnosis (zero overhead when disabled).
- `src/ftl/ftl.c` — retry-budget hardening (~50 LOC).
- `src/hal/hal.c`, `src/media/media.c`, `src/vhost/nbd_async.c` —
  trace points on error returns (env-gated, zero-overhead disabled).

## Spec / plan deltas (acknowledged drift)

- Spec §4.3 "Fix shape table" guesses (block_alloc / L2P update / HAL
  parallel program / writer-vs-GC dst block) were all wrong.
- Plan Phase 3 "<100 LOC FTL only" — final diff is ~50 LOC FTL only,
  matches the constraint. The intermediate cv attempt expanded scope
  but was reverted before merge.
- v1 root cause doc (`...rootcause.md`) is preserved for history but
  superseded by `...rootcause-v2.md`.

## Tracked follow-ups (out of scope here)

- A cv-based wait-on-busy in cmd_engine could remove polling overhead
  and eliminate the retry budget entirely — but needs a separate
  blocking-vs-nonblocking API distinction so legality assertions
  (IS-04, IS-06, SUSPEND tests) can keep using synchronous-BUSY form.
- Reduce reproducer-side MISMATCH counts (verify-ordering artifacts
  under concurrent same-LBA writes — reproducer bookkeeping, not FTL race).
- Optional cleanup of `HFSSS_TRACE_IO_ERR` scaffolding once the bug
  is confidently retired.


AlvinYangZF:
## Spec Review: PR #115 — Async FTL Race Investigation

Excellent RFC. The problem statement, pipeline topology, and phased approach are clear and well-structured. A few observations to strengthen the spec before Phase 1 begins.

---

### Technical Observations

#### 1. The "Fail-Loud" signal is underutilized [MEDIUM]

The spec notes that the production path returns `SCT=0x2 SC=0x80` (write fault) while the existing reproducer shows silent corruption. This is a significant localization clue: `SC=0x80` maps to a specific `return -ERR` statement somewhere in the code — almost certainly a defensive check inside `media_nand_program` or its callers that fires when an illegal state (e.g. double-program on the same physical page) is detected.

**Suggestion:** Before Phase 1, trace `SC=0x80` backward to the exact source line. This may give you the race location directly, potentially collapsing Phase 2 from 2 days to hours.

```
SCT=0x2 SC=0x80 → NVMe INTERNAL_ERROR
  → src/hal/hal_nand.c:media_nand_program return path
  → src/media/nand.c:nand_program_page defensive check
  → which invariant is violated?
```

#### 2. Dispatcher reproducer may shift the Heisenbug [LOW]

Phase 1 builds a new reproducer that mirrors the async pipeline topology. There is a risk that the interleaving produced by the new dispatcher model triggers the race in a slightly different profile than `pre-checkin` does. A fix validated against the Phase 1 reproducer might not actually fix the `pre-checkin` failure (or might fix it at a different rate).

**Suggestion:** Validate the Phase 1 reproducer against the *actual* pre-checkin failure by comparing failure signatures (same NVMe status code? same fio cases? same LBA ranges?) before moving to Phase 2. If the signatures differ, tune the reproducer knobs until they match.

#### 3. Memory pressure from 2048 MiB vs 512 MiB [MEDIUM]

The `pre-checkin` path uses 2048 MiB exported size; `qemu-blackbox` (mt) uses 512 MiB. The larger size means more blocks in the free list, more shard activity, and increased likelihood of hitting boundary conditions in `block_alloc_from_shard` — especially at the shard-wrap boundary or near free-list exhaustion. This is a confounder that the spec doesn't explicitly address.

**Suggestion:** Test the Phase 1 reproducer at both `--export-size=512` and `--export-size=2048` before committing to the dispatcher-topology hypothesis. If the rate is the same at both sizes, memory pressure is not a factor. If it scales with size, the allocator becomes a prime suspect.

#### 4. GC destination block race not in invariant trace table [SUGGESTION]

Phase 2's tuple trace table lists three invariants. A fourth should be added: **GC destination block — at most one host write to a GC destination block while GC is populating it** (complement to `7b5a92e` which covers the *source* page). The fix mapping table in §4.3 already lists this case, but the trace table doesn't.

---

### Answers to Open Questions

| Question | Recommendation | Rationale |
|----------|---------------|-----------|
| **Q1** — Single vs split PR | **Option A** (single PR) | The reproducer is the fix's proof. Two reviews slow momentum for no gain. |
| **Q2** — Reproducer permanence | **Option A** (opt-in make target) | Add `make ftl-mfc-repro-dispatcher` but also wire it into a weekly/nightly CI target if one exists. |
| **Q3** — Escalation rule | **3 days** (not 2) | TSan silence + strategy B setup alone can eat day 1. A 3-day budget gives breathing room before the "cut losses" call. The binary-search mitigation (halve workers until rate drops) suggested in §4.2 is a good intermediate step. |
| **Q4** — Out of scope | **Confirmed** | All four constraints are appropriate. |

---

### Minor Suggestions

- **Phase 1 gate threshold:** The spec says ≥ 1 fault per 1K ops or pivot. What if the rate is ~1 per 10K? That's still a real bug — consider adding an "acceptable" vs "unacceptable" rate definition (e.g. ≥ 1 per 5K passes the gate, < 1 per 5K triggers a lighter investigation before deciding to pivot).
- **p99 ±10% tolerance:** Consider tightening to ±5% for a correctness-only fix (no performance changes expected).
- **Section 5.1 pre-fix baseline:** Explicitly call out that the baseline should include the failure rate on the Phase 1 reproducer AND on `make pre-checkin` (already captured in the e2e audit).

---

### Summary

The spec is solid. The phased approach, risk register, and validation gates are well-thought-out. The three items above (#1 fail-loud trace, #3 memory pressure, #4 GC destination block) would strengthen the investigation plan.

No design-level objections. Happy to approve once the items above are addressed or acknowledged.

AlvinYangZF:
Plan committed: `docs/superpowers/plans/2026-04-26-async-ftl-race.md` (commit `0b13fee`).

Highlights vs spec:

- **Phase 1 refinement** — capture baseline on the *existing* reproducer **first** before adding any new code. Spec §4.1 said the dispatcher-topology hypothesis is testable; the plan makes it the actual test (skip Task 1.2 if existing repro already shows ≥ 1/1K rate at threads=32 / lba=4096).
- **Phase 2** — TSan first per spec §4.2; tracing fallback only if TSan is silent.
- **Phase 3** — hard constraint table at Task 3.1 Step 1 mirrors spec §3.2 non-goals + §4.3 < 100 LOC ceiling.
- **Phase 4** — all five validation lanes covered (make test / make systest / make pre-checkin async / make bench-cq perf / closeout doc).

Default decisions for spec §7 open questions:
| Q | Default |
|---|---|
| Q1 PR structure | single PR (this branch) |
| Q2 reproducer permanence | Phase 4 sub-task adds opt-in Makefile target if useful |
| Q3 phase 2 escalation | 2-day cap, then escalate before any fix |
| Q4 out-of-scope | codified as Phase 3 hard constraints |

Self-review checklist at the bottom of the plan crosses spec §3 / §4 / §6 / §7 against tasks. If you disagree with any Q1-Q4 default, push back before plan approval.

Ready for plan review.

AlvinYangZF:
## Phase 1 closeout (commit `f1c112c`)

Pre-fix baseline locked. **Important spec deviation**: the rate-amplification idea in spec §4.1 turned out to be a wrong frame. The real signal was being silently dropped — `ftl_mt_submit` / `ftl_mt_poll_completion` returned non-OK status that the existing reproducer ignored.

### What now works

```
./build/bin/ftl_mfc_repro --mode legacy --threads 128 \
                          --rwmix 70 --duration 10 --lbas 256

ops=95740 mismatches=26 errors=7818 (write=0 read=7818)
err_rate=7.55e-02
```

7.5% read-error rate — **75x higher than spec's 1/1K target**. Single command, 10 s. Reproducible across runs (3.7%-7.7% range).

### Failure shape

| | rate | type |
|---|---|---|
| **READ status=-6 (HFSSS_ERR_NOENT)** | ~7.5% | the dominant signal |
| WRITE status non-OK | 0 | none observed in reproducer |
| MISMATCH (verify-CRC) | ~3e-4 | partly verify-ordering false positives, kept as cross-check |

`HFSSS_ERR_NOENT` on read = L2P lookup failed (no mapping for the LBA). But the workload writes most LBAs many times. Strongly suggests **L2P-side race**: writer mid-update → reader sees cleared/transient mapping → -6.

### vs pre-checkin

Pre-checkin's NVMe `SCT=0x2 / SC=0x80` is a **WRITE Fault**, while my reproducer triggers **READ NOENT**. Different surfaces of (likely) the same underlying race — both indicate FTL state corruption visible from outside. Phase 2 should investigate both signals; if Phase 2 nails the L2P race the WRITE Fault may resolve as a secondary effect.

### --mode dispatcher kept but deprioritized

Single-producer dispatcher mode wedges under ring back-pressure (producer overruns the SPSC rings faster than workers drain). Not a useful baseline as-is. Phase 2 uses legacy mode.

### Next: Phase 2

Phase 2 starts now per plan: TSan first, tuple-tracing fallback, output a root-cause artifact at `docs/superpowers/artifacts/2026-04-26-async-ftl-race-rootcause.md`. Will commit on this same branch.

AlvinYangZF:
## Phase 2 closeout (commit `10d78f8`)

Root cause confirmed. **It's the second race the 2026-04-18 closeout pre-named.**

### The race

`src/ftl/gc.c:493` and `gc.c:534` — GC's relocation failure branches do unconditional `taa_remove(taa, lba)` even though a host writer may have remapped `lba` to a different PPN since GC sampled `src_ppn` at the top of the loop. The remove wipes the host's fresh mapping, and subsequent host ops on that LBA fail.

Fix `7b5a92e` (CAS for the GC success path) was the right pattern but only applied to the success branch. The failure branches were missed.

### TSan triage (Phase 2.1)

6 unique warnings, all benign stats-counter races (`host_write_pages++`, `read_retry_count++`, `uncorrectable_count++`, harness `total_ops++`/`mismatch_count++`). None explain NOENT. The actual race is logical state-machine ordering, not memory order — TSan structurally cannot catch it.

### Why this matches pre-checkin's NVMe `SCT=0x2 / SC=0x80` (Write Fault)

Same race, different op surface. Reproducer reads dominate (rwmix=70) so the deleted L2P slot surfaces as read-NOENT. Pre-checkin's fio mix and FTL's write path traverse the L2P lookup of the prior PPN to mark it superseded — that lookup fails on a wiped slot and the failure propagates to NVMe as Write Fault.

### Fix shape (Phase 3 — starting now)

Add CAS-style sibling to `taa_update_if_equal`:

```c
int taa_remove_if_equal(struct taa_ctx *ctx, u64 lba,
                        union ppn expected_old, bool *removed_out);
```

Replace the 2 call sites in `gc.c`:

```c
- taa_remove(taa, lba);
+ taa_remove_if_equal(taa, lba, src_ppn, NULL);
```

Estimated < 50 LOC. Well under spec's < 100 LOC ceiling. Touches only `src/ftl/taa.{c,h}` and `src/ftl/gc.c` — no `channel_worker`, no FTL data structure refactor, no HAL timing.

Full root cause + evidence + scope confirmation: [docs/superpowers/artifacts/2026-04-26-async-ftl-race-rootcause.md](https://github.com/AlvinYangZF/hfsss-simulator/blob/feat/async-ftl-race-investigation/docs/superpowers/artifacts/2026-04-26-async-ftl-race-rootcause.md).

AlvinYangZF (commented):
## Review: PR #115 v2 — Async FTL Race Investigation (Path A Pivot)

**Status**: `pre-checkin 8/8 PASS` ✅, but root-cause hypothesis refuted by author — excellent self-correction.

---

### 1. Context Handoff Document Quality [HIGH]

The `2026-04-26-async-ftl-race-context-handoff.md` is **outstanding** engineering writing. Most engineers would bury the "I was wrong" part; you turned it into a navigable document that someone else (or future-you) can read and resume from in 30 minutes. The Path A step-by-step, the "discard these files" list, the quick-restart commands — this is how async investigations should be handed off.

**One suggestion**: Add a "Confidence Level" field to each phase conclusion. E.g.:
```
Phase 1: HIGH — error tracking is real, warmup filter is real
Phase 2 attempt-1: LOW — refuted by taa_remove_if_equal zero-fire + LBA trace
```
This makes the "pivot or persevere" decision visible to reviewers.

---

### 2. The "Two Signals" Problem [CRITICAL — needs resolution]

Current state:
- **Reproducer signal**: READ NOENT on unwritten LBAs → warmup artifact (not a race)
- **Pre-checkin signal**: NVMe Write Fault (SCT=0x2 SC=0x80) on fio-010 — **real bug, not yet localized**

These are **different surfaces**. The risk is that fixing the reproducer's warmup artifact (Step 2 filter) gives you a clean reproducer that **still doesn't reproduce** the pre-checkin failure. Then you're stuck: no reproducer, no fix target.

**Suggestion**: Before discarding the speculative fixes (Path A Step 1), consider running `make pre-checkin` WITH the `taa_remove_if_equal` patch applied. If pre-checkin goes from 8/8 FAIL to 8/8 PASS, then the two signals *do* share a root cause — they just surface differently due to workload mix. If pre-checkin still fails, then the signals are genuinely separate and Path A (direct fio-010 instrumentation) is correct.

This test takes ~30 minutes and saves you from discarding a potentially correct fix.

---

### 3. Step 4a — FIO-010 Instrumentation Strategy [MEDIUM]

Your proposed `fprintf(stderr, ...)` at every `return HFSSS_ERR_*` is good for first-pass localization. But fio-010 generates a lot of I/O; `stderr` tracing will be noisy.

**Suggestion**: Add LBA-specific tracing. Pick the first failing LBA from the fio-010 log, then gate `fprintf` on `lba == target_lba`. This gives you a clean "who touched this LBA and in what order" timeline without drowning in unrelated traffic.

Also: the `IO_ERR` trace log you already have (`2026-04-26-async-ftl-race-fio010-trace.log`) shows `media_nand_program` returning `-4` (likely `HFSSS_ERR_BUSY` or `HFSSS_ERR_INVALID`). This is already a strong signal — the write fault is happening at the HAL/media layer, not the FTL L2P layer. Phase 2's focus on `taa_remove` was a layer-too-high hypothesis.

---

### 4. Commit History Hygiene [SUGGESTION]

You have 13 commits on this branch, including the wrong-root-cause doc (`10d78f8`). Before final merge, consider:
- Squashing the speculative fix commits (`bd7678a`, `0c88d36`) into a single "attempt: FTL retry hardening (intermediate, not the root cause)" commit
- Keeping the wrong-root-cause doc but appending a header: "⚠️ Superseded by [link to new doc]"
- The final fix commit (`2209755`) should have a clear commit message that references the *correct* root cause, not the superseded one

---

### 5. The "GC dst block" item from my v1 review [FOLLOW-UP]

My first review suggested adding "GC destination block — at most one host write" to the invariant trace table. Given that `taa_remove_if_equal` fired **zero times** in the reproducer, this invariant is likely not violated in the current workload. But: is it violated in pre-checkin? The fio-010 trace shows block 511 on chip 3 die 1 — check if this block was a GC destination at the time of the write fault.

This is a 5-minute check (`grep` the trace for block 511 around the fault timestamp, cross-reference with GC logs) that either eliminates or elevates this hypothesis.

---

### Summary

The pivot from "I found the root cause" to "I was wrong and here's the new plan" is the hardest thing in debugging. You've done it well. The PR should remain open as an **investigation** document, not a fix PR, until Step 4a localizes the actual pre-checkin failure.

**Recommended next action**: Run pre-checkin WITH the speculative fix applied (30-min test). If it passes, the signals share a root cause and you can revive the fix. If it fails, proceed with Path A Step 4a.

No design objections to the handoff structure. The spec/plan deviations are properly documented.

---

*Reviewed by: model-router review agent (deepseek-v4-flash)*
*Date: 2026-04-27*

AlvinYangZF:
## ⚠️ Risk Assessment: PR Frame Mismatch

Based on the context handoff doc and investigation timeline, this PR carries **moderate-to-high merge risk** in its current state:

### Risk 1: Title/Content Mismatch 🔴

The PR title claims `fix: async-FTL race`, but the investigation has **refuted the original race hypothesis** (`taa_remove` in `gc.c`). The Phase 2 conclusion was wrong; the actual root cause is **still unknown**.

**Impact**: Merging a PR titled "fix" that doesn't actually fix the claimed bug creates technical debt and confusion for future maintainers.

**Mitigation**: Update title to `investigation: async-FTL race` or `diagnostic: ...` until the true root cause is localized.

### Risk 2: Uncommitted Speculative Fixes 🔴

The author identified 4 uncommitted files (`taa.h`, `taa.c`, `gc.c`, `ftl.c`) that should be **discarded** per Path A Step 1. If these accidentally get committed before the PR merges, the codebase gains dead code that "doesn't touch the actual cause".

**Mitigation**: Verify `git status` is clean before merge; or explicitly exclude these files.

### Risk 3: Two-Signal Confusion 🟡

The reproducer's `NOENT` (warmup artifact) and pre-checkin's `Write Fault` are **treated as separate signals**, but this assumption is **unverified**.

**Impact**: If they share a root cause, discarding the `taa_remove_if_equal` fix is a false negative — we might throw away a correct fix that needs a better reproducer.

**Mitigation**: Run the 30-minute validation test (pre-checkin WITH the speculative fix applied) before finalizing Path A.

### Risk 4: Documentation Drift 🟡

The commit `10d78f8` ("docs(artifact): root cause document") contains a **wrong root cause**. It's marked as "⚠️ WRONG ROOT CAUSE" in the handoff doc, but the commit itself doesn't carry that warning.

**Impact**: Someone reading commit history without the handoff doc may be misled.

**Mitigation**: Amend commit message or add a follow-up commit that superseded it with clear references.

---

### Recommendation

1. **Do NOT merge as-is** with "fix" in the title
2. **Convert to Draft** and reframe as investigation/diagnostic
3. **Complete the 30-min validation** (pre-checkin + speculative fix) to verify signal independence
4. **Land `f1c112c` and `2209755` as separate PRs** if they have independent value
5. **Update or supersede `10d78f8`** with corrected analysis

The engineering rigor here is excellent — the risk is purely in **framing and premature merge**.

---

*Posted by: code review agent (deepseek-v4-flash via model-router)*

PR #116, opened 2026-04-26
fix(pcie): suppress compiler warnings for unused params and operator precedence

AlvinYangZF:
## Summary

Two small fixes to suppress compiler warnings flagged by `-Wall`:

### 1. `dma.c` — mark unused function parameters

`dma_copy_from_prp` and `dma_copy_to_prp` declare `prp1`, `prp2`, and `page_size` parameters that are currently unused (the functions validate the PRP list but actual DMA dispatch is a no-op in the simulated environment). Added `(void)` casts per project convention.

### 2. `pci.c` — add parentheses for operator precedence

`gcc -Wlogical-op-parentheses` flags `size != 1 && size != 2 && size != 4` inside `||` chains in `pci_dev_cfg_read` and `pci_dev_cfg_write`. The behavior is correct (`&&` binds tighter than `||`) but explicit parentheses silence the warning and make the intent unambiguous.

## Verification

- `make -j12` builds cleanly with 0 errors
- No behavioral changes — all fixes are warning-only
- Existing `make test` unaffected

## Before/After

```
Before: 6 warnings from these two files
After:  0 warnings from these two files
```

### DIFF
--- a/src/pcie/dma.c
+++ b/src/pcie/dma.c
@@ -43,6 +43,7 @@ void dma_cleanup(struct dma_ctx *ctx)
 
 int dma_copy_from_prp(struct dma_ctx *ctx, u8 *dst, u64 prp1, u64 prp2, u32 len, u32 page_size)
 {
+    (void)prp1; (void)prp2; (void)page_size;
     if (!ctx || !dst) {
         return HFSSS_ERR_INVAL;
     }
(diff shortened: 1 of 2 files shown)


AlvinYangZF (commented):
## Review: PR #116 — fix(pcie): suppress compiler warnings

**Status**: CI pass ✅ | Simple, low-risk fix

### dma.c: (void) cast for unused params ✅

Standard idiom, zero runtime overhead. No issues.

### pci.c: Operator precedence parentheses ✅

The original expression:
```c
if (!dev || !value || size != 1 && size != 2 && size != 4)
```

With C precedence (`&&`=11, `||`=12), this parses correctly as:
```c
(!dev) || (!value) || ((size != 1) && (size != 2) && (size != 4))
```

So **this is not a logic bug** — the original semantics were already correct. The fix suppresses the `-Wparentheses` warning by making the grouping explicit. This is the standard recommended practice and matches the PR title.

### Final Verdict: ✅ APPROVE — ready to merge

Minimal, correct, CI passes. Recommend squash merge into a single commit.

PR #118, opened 2026-05-08
perf(ftl): per-die wait-queue + completion-driven dispatch (correctness only; latency follow-up needed)

AlvinYangZF:
## Summary

- Replace FTL's `nanosleep × 2048` retry-spin with a per-die wait queue + completion-event-driven dispatcher (`src/ftl/die_dispatcher.c`).
- Add multi-tier priority scheduling (T1 critical GC > T2 host_read+force_GC WFQ > T3 host_write+normal_GC WFQ > T4 idle GC with host-burst yield via EMA).
- Thread `gc_trigger_t` through the GC pipeline so GC submits at the appropriate priority class.
- New `cmd_engine` notifier hook fires when a die transitions to `DIE_IDLE` (anchors A and B; cache-implicit-IDLE at submit entry deliberately skipped). Default-NULL hook preserves byte-identical behavior for every existing caller that does not install a dispatcher.
- Six-layer test coverage: L1 pure data structure (25 cases), L2 multi-thread w/ mocked cmd_engine (11), L3 real cmd_engine integration (22), L4 priority + WFQ (5 statistical scenarios), L5 regression (existing red line preserved), L6 end-to-end pre-checkin (8/8 PASS).
- Two env-gated stress knobs: `HFSSS_DIE_DISP_FORCE_BUSY=N` and `HFSSS_DIE_DISP_NOTIFIER_DELAY_NS=K`.

## Honest perf reporting

The original spec target was 012_fio_seqwrite_verify mean 142 ms → ≤ 10 ms. **This PR does not deliver that.** Observed: 141.8 ms mean (essentially identical to baseline). The p99 trims slightly from 152.0 ms to 143.6 ms.

Root cause: `cmd_engine`'s `die_lock` and `channel_lock` are `pthread_mutex_t`, which is not FIFO-fair. The dispatcher's wake order is undone by lock races against fresh arrivals.

## Follow-up plan (committed in this PR for traceability)

The smallest scoped follow-up that actually closes the latency gap is committed alongside this PR as documentation only:

- Spec: `docs/superpowers/specs/2026-05-08-cmd-engine-ticket-lock-design.md`
- Plan: `docs/superpowers/plans/2026-05-08-cmd-engine-ticket-lock.md`

The plan replaces `pthread_mutex_t` for cmd_engine's two locks with a FIFO-fair ticket lock primitive (~30 LOC + tests). Execution starts on `perf/cmd-engine-ticket-lock` after this PR merges. Two of the three options in Section 9.2 of the dispatcher spec are deferred (die-reservation protocol, REQ-045 tier 3 channel_worker retarget) — both larger architectural changes that build on top of either the dispatcher or the ticket-lock foundation.

## What this PR does deliver

- **Correctness under fio-012**: passes 8/8 pre-checkin (was failing on the master fix path under stress without the budget bump).
- **Architectural foundation**: per-die wait queue + completion-event-driven wakeup is the canonical real-SSD-controller pattern.
- **Priority infrastructure**: GC origin tagging is now plumbed end to end.
- **Stress hooks**: two env knobs let CI exercise the wake/retry path under controlled fault injection.
- **Documented next step**: the ticket-lock follow-up has spec + plan ready to execute.

## Test plan
- [x] `make test` — full suite green
- [x] `make pre-checkin` — 8/8 PASS, 1 SKIP (900_spdk by design)
- [x] L1 pure data structure: 25/25
- [x] L2 multi-thread mocked engine: 11/11 (3-run stable, 170k waits + 240k notifies under stress)
- [x] L3 real cmd_engine integration: 22/22
- [x] L4 priority + WFQ scenarios: 5 cases, 3-run stable
- [x] Regression red line: `test_media` IS-04 / IS-06 / SUSPENDED, `test_cmd_engine_notifier`, `test_channel_worker`, `test_reset_abort_race` — all PASS
- [x] Baseline archived: `docs/perf-baselines/2026-04-29-master-0eb0d83/` and `docs/perf-baselines/2026-04-30-perf-ftl-die-busy-waitqueue/`
- [ ] Reviewer to decide: merge as foundational infrastructure, follow-up ticket lock then closes the latency gap as documented

## Spec / plan
- Dispatcher spec: `docs/superpowers/specs/2026-04-30-ftl-die-busy-waitqueue-design.md`
- Dispatcher plan: `docs/superpowers/plans/2026-04-30-ftl-die-busy-waitqueue.md`
- Ticket-lock follow-up spec: `docs/superpowers/specs/2026-05-08-cmd-engine-ticket-lock-design.md`
- Ticket-lock follow-up plan: `docs/superpowers/plans/2026-05-08-cmd-engine-ticket-lock.md`


AlvinYangZF:
## Review: `perf(ftl): per-die wait-queue + completion-driven dispatch`

@AlvinYangZF — overall very solid architecture. The per-die wait queue + priority-tiered wakeup + EMA burst detection is the correct design for an SSD simulator moving beyond nanosleep retry-spin. Review focused on concurrency correctness, integration safety, and test coverage.

---

### CRITICAL: `rand()` is not thread-safe

**File:** `src/ftl/die_dispatcher.c` ~line 610

```c
if (pct > 0 && (rand() % 100) < pct) {
```

`rand()` uses global libc state and is explicitly **not thread-safe** (POSIX lists it as MT-Unsafe). When multiple FTL worker threads concurrently call `die_dispatcher_wait`, they will corrupt each other's RNG state. On glibc this produces garbled sequences; on macOS (the dev platform) even worse — data races on the global seed.

**Fix:** Replace with a relaxed atomic counter + modulo:

```c
static _Atomic u32 g_inject_counter;
// ...
u32 c = atomic_fetch_add_explicit(&g_inject_counter, 1, memory_order_relaxed);
if (pct > 0 && (c % 100) < (u32)pct) {
```

This also eliminates the hidden lock inside glibc's `rand()` which would serialize all dispatcher waiters and negate the parallelism the wait-queue design achieves.

---

### IMPORTANT

**1. Memory ordering on env-cache reset (ARM target)**

`die_dispatcher_reset_env_cache_for_testing` stores `-1` with `memory_order_relaxed`. The project targets `aarch64` (QEMU `virt,gic-version=3`), where relaxed stores have no ordering guarantees. A reader thread in `die_disp_force_busy_pct()` doing a relaxed load may never observe the cache reset if the test thread's store stays in the store buffer.

This only affects test code (production never calls reset), so it's not a release bug — but CI `make test` runs on ARM macOS, and a flaky test caused by this would be very confusing. Recommend upgrading to `memory_order_release` on the store and `memory_order_acquire` on the load, or documenting that tests calling `reset_env_cache_for_testing` must do so before spawning dispatcher threads.

**2. GC retry budget divergence from host-IO budget**

Host IO: `FTL_DISPATCH_SAFETY_RETRIES = 2048 × 10ms = 20s`
GC: `GC_DISPATCH_SAFETY_RETRIES = 8 × 50ms = 400ms`

The 50× ratio is explained in the spec (GC runs inline under TAA; a hung GC pass blocks host writes). But if the dispatcher itself is broken or the die is genuinely stuck, GC exhausts its budget in 400ms and returns the last error, which may cascade to host IO returning EIO/Write Fault. This is the same failure mode the PR is fixing on the host side. Should GC inherit the host budget, or is the 400ms ceiling intentional to bound GC's blocking window on host IO?

If intentional — worth a comment in gc.c justifying the 400ms bound.

**3. `dispatch_list_del` conflates "not in list" with "already dequeued"**

The `die_dispatcher_wait` timeout path uses `dispatch_list_empty(&w.list)` to detect whether the notifier already popped the waiter. This works because `dispatch_list_del` makes the node self-looping, and `dispatch_list_empty` on a self-loop returns true. But this relies on the invariant that a newly-initialized waiter also starts self-looping (`dispatch_list_init(&w.list)` in `die_waiter_init`). If anyone ever changes `die_waiter_init` without understanding this contract, the timeout path would double-delete.

Consider adding an explicit state enum (`WAITING`, `DEQUEUED`, `TIMED_OUT`) to the waiter struct instead of overloading the list link state. Not blocking — just a maintainability concern for a 634-line file that will grow.

---

### What this PR does well

- **Correct shutdown semantics**: `die_dispatcher_destroy` uninstalls the notifier first, then drains every queue with `shutdown=true` under `q->lock`. The waiter observes `shutdown` before touching any freed memory. Clean.
- **Backward compatibility**: Default-NULL notifier hook is genuinely zero-overhead. `nand_device` callers that don't install a dispatcher pay one NULL-check branch per die transition.
- **Test layering**: L1–L4 progress from pure data structure → mocked multi-thread → real engine integration → statistical WFQ validation. The mock (`cmd_engine_mock.c`) faithfully reproduces the single observable effect the real engine has on the dispatcher.
- **Honest perf reporting**: The PR description explicitly states the latency target was NOT met (142ms → 142ms) and identifies the root cause (pthread mutex unfairness defeating wake order) with a committed follow-up plan. This is the kind of integrity that makes code review easier.
- **Env-gated fault injection**: `HFSSS_DIE_DISP_FORCE_BUSY` and `HFSSS_DIE_DISP_NOTIFIER_DELAY_NS` are well-designed — zero cost when unset (single relaxed load), and they test the re-queue and widening-window paths that are otherwise hard to reach.

---

### NIT

- `gc_hal_read_with_wait` / `gc_hal_program_with_wait` / `gc_hal_erase_with_wait` share identical retry-loop structure. Consider extracting a shared helper to reduce the ~90 lines of copy-paste.
- `die_priority_slot(DIE_PRIO_GC_CRITICAL)` returns `-1` — unreachable today but would silently insert into wrong WFQ slot if T1 ever moved to a shared tier.
- `strtoull` in `die_disp_notifier_delay_ns` doesn't check `errno` — bad input silently becomes 0. Conservative but a `HFSSS_LOG_WARN` on parse failure would save debugging time.

---

### Verdict

**APPROVE** with one required fix: replace `rand()` with thread-safe alternative before merge.

The architecture is sound. The performance gap is understood, documented, and has a committed follow-up. The test coverage is thorough. The `rand()` issue is the only correctness bug I found that can manifest in production (multi-threaded FTL under `fio` stress).

Nice work on the shutdown sequence — that's where most wait-queue implementations have bugs, and this one is clean.

AlvinYangZF:
## Review response

Pushed `db78cea` (or whatever HEAD is post-push) addressing the review.

### CRITICAL (fixed)
- **`rand()` → atomic counter** in the force-BUSY injection. `g_force_busy_counter` is a `_Atomic uint32_t` incremented per wake via `atomic_fetch_add_explicit(..., memory_order_relaxed)`. Statistically equivalent for uniform injection rates and removes the libc RNG hidden serialization. `src/ftl/die_dispatcher.c`.

### IMPORTANT (fixed)
1. **Memory ordering**: `die_disp_force_busy_pct` and `die_disp_notifier_delay_ns` now use `memory_order_acquire` on the load + `memory_order_release` on the store. `die_dispatcher_reset_env_cache_for_testing` matches with `memory_order_release`. Comment makes the test-thread-must-quiesce-before-spawning expectation explicit.
2. **GC retry budget rationale**: added a comment block above `GC_DISPATCH_SAFETY_RETRIES` in `src/ftl/gc.c` explaining the 400 ms ceiling vs the host-IO 20 s — GC runs inline from the host write path under NOSPC, so absorbing the full NVMe timeout would stall every host write; the recoverable failure mode (return BUSY → host retry) is the intended bound.
3. **`dispatch_list_del` "not in list" overload**: noted; not addressed in this fix-up. The contract is one place (`die_waiter_init` → `dispatch_list_init` self-loop) and changing it without updating the timeout path would fail loudly under existing tests, so I left it alone for now. If it grows fragile in the ticket-lock follow-up I'll add the explicit state enum then.

### NIT (partial)
- **`strtoull` errno check**: fixed. Bad input now caches as 0 explicitly via `errno` + end-pointer check.
- **`gc_hal_*` retry-loop duplication**: noted, not addressed — extracting a shared helper here would cross the FTL/HAL boundary in ways the current layering avoids; better as a separate cleanup PR if it persists past the ticket-lock work.
- **`die_priority_slot(DIE_PRIO_GC_CRITICAL)` returns -1**: noted — the current callers all pre-tier-check before invoking `die_priority_slot`, so the -1 is unreachable; documenting the precondition as a comment is the lighter-weight option.

Verification: full `make test` exit 0, all dispatcher + regression suites pass byte-equivalent (test_die_dispatcher_unit 25/25, _mt 11/11, _engine 22/22, _faults 6/6, test_cmd_engine_notifier 13/13, test_media 489/489, test_gc_mt 24/24).

Thank you for the thorough review — the `rand()` thread-safety in particular was a real bug that fio-012's 8-worker pressure would have hit.

PR #127, opened 2026-05-23
feat(ftl): implement data superblock write, GC, and checkpoint flows

AlvinYangZF:
## Summary

- Add data superblock allocation, striping write placement, state transitions, and DSB-aware accounting.
- Route FTL host writes through DSB placement and support DSB-aware direct/MT GC victim flow.
- Persist DSB metadata through superblock checkpoint v2, restore open/closed/GC DSB state on recovery, and compact multi-LBA TRIM into range journal entries.
- Harden UT coverage capture so test-only gcda/gcno path mismatches do not abort report generation.

## Tests

- `make test`
- `make coverage-clean`
- `make coverage-ut`
  - lines: 85.1% (15203 / 17870)
  - functions: 95.0% (1080 / 1137)
  - branches: 67.3% (6586 / 9784)
  - report: `build-cov/coverage/ut/index.html`
- `bash scripts/coverage/ratchet_check.sh`
- `git diff --check`

PR #128, opened 2026-05-23
[codex] add FTL refactor and dump rebuild plans

AlvinYangZF:
## Summary

Adds two standalone Superpowers implementation plans:

- `2026-05-23-single-ftl-read-write-split-dsb.md`
  - Refactors the FTL data path into one authoritative FTL core.
  - Replaces LBA-hash worker ownership with write/TRIM mutation worker plus scalable read workers.
  - Designs runtime L2P/P2L, mapping mutation log, TRIM log, multi-stream placement, multi-open DSB, backend scheduler execution, and debug counters.
  - Explicitly excludes DSB internal bad-block/layout/retire policy and power/checkpoint/recovery changes.

- `2026-05-23-ftl-dump-rebuild.md`
  - Designs an independent dump/rebuild mechanism for L2P, multi-level mapping pages, map logs, journal mirrors, TRIM bitmap/range state, and derived P2L.
  - Keeps P2L as debug/derived state, not persistent recovery authority.
  - Explicitly excludes normal power-on recovery and UPLP integration.

## Validation

- `git diff --cached --check`
- `git diff --check origin/master...HEAD`

No source code changes are included.


AlvinYangZF:
DeepSeek CLI review completed for this merged docs PR. It found several potential follow-up risks worth tightening before these plans are used as execution instructions:

1. `docs/superpowers/plans/2026-05-23-single-ftl-read-write-split-dsb.md` references DSB/backend-scheduler artifacts such as `backend_sched_init` and `build/bin/test_backend_sched` around lines 492-512, 1346, and 1510. The plan does mention PR #127 as a reference, but the prerequisite is easy to miss now that this document is merged into `master`. Consider adding an explicit prerequisite block: this plan must be executed on top of `codex/data-superblock-phase1` or any branch that already contains DSB and backend scheduler APIs/tests.

2. `docs/superpowers/plans/2026-05-23-ftl-dump-rebuild.md` defines `crc32` in `ftl_dump_section_header` and `header_crc32` in `ftl_dump_file_header`, but does not precisely define the CRC byte ranges. Please specify whether section CRC covers payload only or header-with-crc-zeroed plus payload, and how `header_crc32` is computed. Without this, independent writer/reader implementations can become incompatible.

3. The dump plan defines `FTL_DUMP_MAP_PAGE_ENTRIES = 512` while also carrying `entry_count` in `ftl_dump_map_page_record` around lines 561-588. Please clarify the on-disk serialization rule: always write all 512 entries, or write only `entry_count` entries. Otherwise reader/writer length calculation can diverge.

4. The live dump plan locks `ctx->lock` while copying L2P/journal records around lines 1231-1240. For large capacities, a flat 4KB-granularity L2P snapshot can be very large, so this can block I/O submission for a long time. Consider specifying a bounded lock strategy, such as pointer/page snapshot under lock followed by async serialization, or explicitly limiting this live dump path to debug/small-capacity use.

5. The dump/rebuild and single-FTL split plans should cross-reference their ordering/dependency more explicitly. The split plan introduces per-stream open DSB state; any later live debug dump that intends to capture runtime placement/debug state should state whether per-stream open DSB pointers are included, excluded, or intentionally derived elsewhere.

